PR #35, opened 2015-10-24
support different project structure and support multi-manifest meger

lansign:
support different project structure and support multi manifest files merge


Piasy:
@lansign Thanks for your pr! I'll check it soon.

Piasy:
@lansign Two questions:
1. source set could be obtained through `project.sourceSets`, why only take the first one? and can sourceSets dir not be `project.projectDir`'s child dir? in other words, is `ProjectHelper.getRelativePathBy` necessary?
2. AndroidManifestRule doesn't support BUCK's rule fully, aar dependency and android library module's manifest should be merged but not, aar dependency's manifest could be merged by adding it into AndroidManifestRule's deps part, but I haven't find out how to merge android library module's manifest now...

PR #69, opened 2016-04-20
Add a "buckProjects" configuration to the OkBuckExtension

kageiit:
- Allows optionally only listing a set of projects to convert to buck. There are cases when some projects should not be converted to buck.
- Also creates a symlink 'plugin' to buildSrc to allow `./gradlew plugin:publishToMavenLocal` for local testing of changes
- Only sets properties from `bintray.properties` file if it exists

### DIFF
--- a/README.md
+++ b/README.md
@@ -130,6 +130,7 @@ okbuck {
                     'armeabi',
             ]
     ]
+    buckProjects = project.subprojects
 }
 ```
 
@@ -154,6 +155,8 @@ you need, default is empty, which will generate all flavors BUCK file
 +  `cpuFilters` is a map, used for controlling BUCK only create the specific cpu architecture
 native library folder in your apk file, the same as `ndk.abiFilter` of gradle, support values
 are: `armeabi`, `armeabi-v7a`, `x86`, `x86_64`, `mips`
++  `buckProjects` is a set of projects to generate buck configs for.
+Default is all sub projects of root project.
 
 ## Troubleshooting
 If you come with bugs of OkBuck, please [open an issue](https://github.com/Piasy/OkBuck/issues/new), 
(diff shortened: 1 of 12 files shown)


Piasy:
@kageiit Thanks for your contribution! But I don't see where to exclude sub modules out, and actually I don't understand why we need exclude some modules out, could you please give me more explanation?

kageiit:
Its more of a "include" instead of exclude.

A perfect example would be having multiple apps/modules in a repo some of which are not compatible with buck maybe due to R.id issue or they have to use "-" in their project name.

This will let anyone to atleast use buck for the apps/modules that are compatible and slowly convert the rest of them over to make them compatible with buck.

I will fix the build issue. its happening because of missing bintray.properties. maybe can just add an empty one

Piasy:
You convinced me. 

I have another question(I'm not quite good at groovy), you set `toBuck = project.subprojects` in extension's constructor, wouldn't `toBuck` always be all sub project? Or when we set it in build.gradle, it will rewrite the default value?

kageiit:
I will probably rename toBuck to buckProjects when i fix this build. Its a better name

toBuck = project.subprojects is the default behavior. i.e it analyzes all subprojects the way you have it right now. But if user chooses to change that, they can set it to rewrite the default value

kageiit:
In addition, this change itself showcases the use case for the `buckProjects` option. The buildSrc (which is now mirrored via `plugin`) is a groovy project. But all groovy projects also apply the java plugin internally by default.

OkBuck will try to process groovy project and fail. This is another valid use case when in repositories, there can be multiple projects which have groovy/scala sub projects in addition to regular android sub projects and we need to specify which projects OkBuck needs to run on.

Piasy:
Great job! Thanks!

PR #70, opened 2016-04-21
Fix OkBuck to work with projects with '-' in their name

### LINKED ISSUE #68: Okbuck fails to process projects with "-" in their name
Easy way to reproduce is to rename "anotherapp" to "another-app" and try to run `./gradlew okbuck` on master

kageiit:
- Uses the Android Gradle Plugin DSL to extract the variant and flavor names by matching the output file of variant to the module dependency output file.
- Test it out by renaming `anotherapp` to `another-app` to prove it works
- Resolves #68


Piasy:
Great job!

PR #71, opened 2016-04-22
Project cleanup

kageiit:
- Updated gradle/dependencies/plugins/buildtools to latest
- Simplified some classes by using groovier syntax
- Removed unnecessary helpers
- Simplified some existing helpers


README.md
```diff
@@ -174,9 +174,8 @@ Maybe there are more caveats waiting for you, but for the super fast build broug
 The rest modules in this repo is a full example usage of OkBuck and BUCK.
 
 ## Compatibility
-OkBuck is tested under `gradle` 2.2.1 ~ 2.10, and `com.android.tools.build:gradle` 1.5.0. 
-Other versions will not considered compatible, if you have compatibility problem, please 
-fire an issue.
+OkBuck is tested under `gradle` 2.12, and `com.android.tools.build:gradle` 2.0.0.
+If other versions have compatibility problems, please file an issue.
```
Piasy: We need test compatibility.

kageiit: I tested it on a moderately large app using gradle 2.12 and android gradle plugin 2.0.0

I also updated all the sample/example projects to use these versions to verify

Piasy: I mean compatibility on older gradle version and android-gradle-plugin version, we should not force developer use the latest version.

kageiit: Of course. I can change the note. This is not forcing it. It still works with versions 1.5.0 to 2.0.0 of gradle android plugin

and 2.2.1 to 2.12 of gradle version

kageiit:
@Piasy The travis check seems to fail compiling groovy for this PR.

It is weird because the same travis command I run locally passes. Can you just check what you are seeing on your machine if possible? I will try to debug in the meanwhile

Piasy:
No problem.

Piasy:
@kageiit My local build also succeed.

README.md
Piasy: Great!

kageiit:
Probably. I have updated the travis script to use jdk8

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -3,4 +3,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-2.10-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-2.12-all.zip
```
Piasy: We also need update gradle version in `buildSrc/gradle/wrapper/gradle-wrapper.properties`.

kageiit: interesting. ok will change it

Piasy:
@kageiit Maybe we should refer `DependencyType` in subclass of `Dependency` by `Dependency.DependencyType` ?

kageiit:
Ok trying that

Piasy:
And [this line](https://github.com/Piasy/OkBuck/blob/master/buildSrc%2Fbuild.gradle#L73):
```gradle
compile 'com.android.tools.build:gradle:1.5.0'
```

We should update to 2.0.0 too.

kageiit:
yep its already updated as part of the PR

Piasy:
@kageiit I think the root cause is the way we refer to `DependencyType`. We should update reference in all subclass of `Dependency`, there are 4.

kageiit:
I think we need to only update in 2 places i.e where we do not inherit from Dependency and inherit from FileDependency instead. I have tried in my latest commit.

kageiit:
Atleast it compiles groovy now. Travis box is unable to find build tools version 23.0.3 now :)

kageiit:
Seems like a known issue with 23.0.3 https://github.com/travis-ci/travis-ci/issues/5891
I will revert just that part and update it once travis is ready. but it works locally with 23.0.3

kageiit:
Finally green! Feel free to merge. I will open a PR for cleaning up the duplicate detection next :)

Piasy:
Great job!

PR #73, opened 2016-05-08
Plugin architecture refresh

kageiit:
This is a significant refactor of the plugin architecture. Some highlights
- Update gradle wrapper and android gradle plugin dependencies to latest
- Many configuration options are computed automatically now
  - resPackages
  - appClassSource
  - cpuFilters
  - flavorFilter (removed)
  - checkDepConflict (removed)
- The okbuck task no longer depends on any project tasks like jar or bundleRelease. Which means the time to create generated BUCK files is only 2-3 seconds + project configuration time
- Full support for all buildType + flavor combinations (not just debug and release)
- The number of tasks generated for buck to compute has gone down by about 25%, so buck builds faster as well
- Manifest merging is now done by the gradle manifest merger during BUCK file generation
- Plugin allows for flavor/buildType/main sourceSets and resources to be inherited and modified in the same way the gradle plugin does.
- Extension configuration options can be fine tuned to apply for the entire app project or just a buildType or just a flavor or for a specific variant (buildType+flavor)
- Allows for project dependencies to be declared under apt/provided configurations
- Cleaned up lot of cruft and unused code/buck rules
- Made codebase more groovy and less java :)
- The new plugin architecture allows for integrating new build systems like Bazel or Pants similar to buck. Support for new build rules will be added in soon.

The chief aspect of this architecture is that it operates upon 'Targets' which are basically build targets for any generic build system that uses DAG model of inputs and outputs to build and resolve dependencies.

TODO: I will be adding in/fixing documentation after this merges. Did not want to make this PR any larger than it already is

Note: I reverted the retrolambda commit temporarily to migrate to the new architecture. Will put back retrolambda support once the refactor merges


buildSrc/src/main/groovy/com/github/piasy/okbuck/model/Target.groovy
```diff
@@ -0,0 +1,178 @@
...
+    Set<String> getAnnotationProcessors() {
+        Set<String> annotationProcessors = [] as Set
+        aptDeps.each { String aptDep ->
+            JarFile jar = new JarFile(new File(aptDep))
+            List aptEntries = jar.entries().findAll { JarEntry entry ->
+                entry.name.equals(PROCESSOR_ENTRY)
+            }
+            if (!aptEntries.empty) {
+                aptEntries.each { JarEntry aptEntry ->
+                    annotationProcessors.add(IOUtils.toString(jar.getInputStream(aptEntry)).trim())
```
Piasy: There are libraries declared multiple annotation processors inside its `META-INF/services/javax.annotation.processing.Processor` file, such as `com.github.frankiesardo:auto-parcel-processor:0.3.1`, this should work well:

```groovy
String[] processors = IOUtils.toString(jar.getInputStream(aptEntry)).trim().split("\\n")
processors.each { String processor ->
       annotationProcessors.add(processor)
}
```

buildSrc/src/main/groovy/com/github/piasy/okbuck/generator/BuckFileGenerator.groovy
```diff
@@ -24,321 +24,166 @@
...
+        androidLibRules.addAll(target.resources.collect { AndroidTarget.ResBundle resBundle ->
+            AndroidResourceRuleComposer.compose(target, resBundle)
+        })
+
+        // BuildConfig
+        androidLibRules.add(AndroidBuildConfigRuleComposer.compose(target))
+
+        // Apt
+        List<String> aptDeps = []
+        String aptRuleName = null
```
Piasy: This variable is not used.

PR #77, opened 2016-05-10
add retrolambda support again

Piasy:


buildSrc/src/main/groovy/com/github/piasy/okbuck/config/RetroLambdaShFile.groovy
```diff
@@ -0,0 +1,53 @@
...
+    public RetroLambdaShFile(String classpath, String buildTarget, String retroLambdaJarFileName) {
+        mClasspath = classpath
+        mBuildTarget = buildTarget
+        mRetroLambdaJarFileName = retroLambdaJarFileName
+    }
+
+    @Override
+    public final void print(PrintStream printer) {
+        printer.println("#!/bin/bash")
+        String androidJarPath = System.getenv("ANDROID_HOME") + File.separator + "platforms" +
```
kageiit: this assumption is not always valid. not everyone has ANDROID_HOME set

Piasy: it's required by buck. buck installation guide required user to set this variable.

kageiit:
I believe we can clean this up a lot. Ill commit on top of this branch. thanks for the initial idea!

buildSrc/src/main/groovy/com/github/piasy/okbuck/config/RetroLambdaShFile.groovy
kageiit: I know it needs it. But this is not the right way to do it as user may not have ANDROID_HOME environment variable set always.

I will post the more reliable method on top of this branch.

PR #89, opened 2016-05-14
architecture clean up

Piasy:


buildSrc/src/main/groovy/com/github/piasy/okbuck/model/AndroidTarget.groovy
```diff
@@ -192,6 +177,10 @@ abstract class AndroidTarget extends JavaLibTarget {
         return FileUtil.getRelativePath(project.projectDir, mergedManifest)
     }
 
+    protected void manipulateManifest(GPathResult manifest) {
```
kageiit: Make this abstract

PR #90, opened 2016-05-17
Proguard Support

### LINKED ISSUE #82: Proguard Support

kageiit:
Resolves #82


Piasy:
In cases user add new code that need change proguard configuration, they should execute okbuck task again, right?

kageiit:
Yes. I will add to docs when we are ready to release

PR #102, opened 2016-05-24
fix: RetroLambda classpath bug

Piasy:


kageiit:
LGTM. merge after travis passes

PR #104, opened 2016-05-25
add debuggable attr into merged manifest

Piasy:
add debuggable attr into merged manifest;
avoid call abstract method inside constructor;

### DIFF
--- a/buildSrc/src/main/groovy/com/github/okbuilds/core/model/AndroidAppTarget.groovy
+++ b/buildSrc/src/main/groovy/com/github/okbuilds/core/model/AndroidAppTarget.groovy
@@ -31,7 +31,7 @@ class AndroidAppTarget extends AndroidLibTarget {
     final int linearAllocHardLimit
     final Set<String> primaryDexPatterns
     final Set<String> exoPackageDependencies
-    final String appClass
+    String appClass
 
     final boolean minifyEnabled
 
@@ -52,7 +52,6 @@ class AndroidAppTarget extends AndroidLibTarget {
         primaryDexPatterns = getProp(okbuck.primaryDexPatterns, []) as Set
         exoPackageDependencies = getProp(okbuck.appLibDependencies, []) as Set
         linearAllocHardLimit = getProp(okbuck.linearAllocHardLimit, DEFAULT_LINEARALLOC_LIMIT) as Integer
-        appClass = extractAppClass()
 
         minifyEnabled = baseVariant.buildType.minifyEnabled
     }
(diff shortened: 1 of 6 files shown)


kageiit:
I don't understand the motivation here. Can you give brief explanation?

buildSrc/src/main/groovy/com/github/okbuilds/core/model/AndroidAppTarget.groovy
```diff
@@ -31,7 +31,7 @@ class AndroidAppTarget extends AndroidLibTarget {
     final int linearAllocHardLimit
     final Set<String> primaryDexPatterns
     final Set<String> exoPackageDependencies
-    final String appClass
+    String appClass
```
kageiit: What is the need to move this?

buildSrc/src/main/groovy/com/github/okbuilds/core/model/AndroidTarget.groovy
```diff
@@ -125,7 +124,7 @@ abstract class AndroidTarget extends JavaLibTarget {
         }.flatten() as Set<String>
     }
 
-    protected String extractMergedManifest() {
+    public AndroidTarget extractMergedManifest() {
```
kageiit: This does not make sense to me. Why is this returning this object instead of a string?

buildSrc/src/main/groovy/com/github/okbuilds/core/util/ProjectUtil.groovy
```diff
@@ -67,12 +67,12 @@ final class ProjectUtil {
         switch (type) {
             case ProjectType.ANDROID_APP:
                 project.android.applicationVariants.collectEntries { BaseVariant variant ->
-                    [variant.name, new AndroidAppTarget(project, variant.name)]
+                    [variant.name, new AndroidAppTarget(project, variant.name).extractMergedManifest()]
```
kageiit: This seems pretty unclean. Maybe there is a simpler way to do this

Piasy:
Because the constructor of `AndroidTarget` is executed before `AndroidAppTarget`'s, so if we write `debuggable`'s value at  `AndroidAppTarget`'s constructor, when `manipulateManifest` is called by `extractMergedManifest`, which is called by `AndroidTarget`'s constructor, we will always see `false`, and besides, we call a abstract method `manipulateManifest` as constructor, which is I remember some book said we should not do it.

buildSrc/src/main/groovy/com/github/okbuilds/core/util/ProjectUtil.groovy
Piasy: Yeah, but I can't find a better way :(

kageiit:
- Use the manifest merger correctly
- Fix applicationIdSuffix not outputting correctly to BuildConfig
- Preserve debuggable attribute to manifest

buildSrc/src/main/groovy/com/github/okbuilds/core/model/AndroidAppTarget.groovy
```diff
@@ -77,20 +79,41 @@ class AndroidAppTarget extends AndroidLibTarget {
...
+    protected String extractMergedManifest() {
+        String mergedManifest = super.extractMergedManifest()
+
+        XmlSlurper slurper = new XmlSlurper()
+        GPathResult manifestXml = slurper.parse(project.file(mergedManifest))
+
+        manifestXml.@package = applicationId + applicationIdSuffix
+        manifestXml.@'android:versionCode' = String.valueOf(versionCode)
+        manifestXml.@'android:versionName' = versionName
+        manifestXml.@'android:debuggable' = String.valueOf(debuggable)
```
Piasy: will this get the right value?

PR #109, opened 2016-05-30
Install buck via gradle for travis

kageiit:


Piasy:
in current directory, `buck@ -> ~/.gradle/caches/okbuilds/buck/bin/buck`, but in `~/.gradle/caches/okbuilds/buck`, it only contains a `.git` directory, nothing else, so travis can't find `buck` executable.

btw, could we ignore the `buck` link in pwd?

kageiit:
- Simplified installation process
- Added waitFor to ensure git index does not get locked between git commands
- Make sure switching repositories and branches works

Piasy:
Still fail.

btw, if we have run `buckInstall` task before, when I re-run it, would it rebuild buck binary?

kageiit:
> btw, if we have run buckInstall task before, when I re-run it, would it rebuild buck binary?

yes it will. I deleted travis cache and rebuild it now to check one final time

kageiit:
abandoning this. will open another

PR #116, opened 2016-06-13
Support provided/compile only configurations

### LINKED ISSUE #62: Provided dependencies not picked up by buck
Hi,

I don't know if this is a buck or okbuck issue, but when I use project lombok or hrisey in my android project, they are only used during compilation. They are typically defined as shown below: -

dependencies {
    provided 'pl.mg6.hrisey:hrisey:0.3.+'
}

Unfortunately when dependencies are set as provided, they give a compilation error when running buck install. Could it be a missing configuration during okbuck generation?

kageiit:
Resolves #62 

These dependencies are not exported and are only used to compile the target similar to gradle behavior


raviagarwal7:
Looks good!

PR #118, opened 2016-06-14
Use buck's provided_deps

kageiit:
- This uses the officially supported `provided_deps` from buck instead of the classpath hack we have currently
- It is not needed for java targets as the `provided_deps` and `deps` are the same


raviagarwal7:
Looks good!

PR #122, opened 2016-06-29
Support provided_deps for java rules

kageiit:


buildSrc/src/main/groovy/com/github/okbuilds/okbuck/rule/JavaRule.groovy
```diff
@@ -50,6 +52,14 @@ abstract class JavaRule extends BuckRule {
             }
         }
 
+        if (!mProvidedDeps.empty) {
```
raviagarwal7: i think we can just have `mProvidedDeps` it checks for both null and empty.

raviagarwal7:
left 1 nit comment. 

LGTM!

buildSrc/src/main/groovy/com/github/okbuilds/okbuck/rule/JavaRule.groovy
kageiit: will change in the coming up cleanup PR

PR #124, opened 2016-06-30
SqlDelight Support

Piasy:
see wiki page for detail: https://github.com/OkBuilds/OkBuck/wiki/Known-caveats#sqldelight-work-around

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -2,6 +2,7 @@ apply plugin: 'com.android.application'
 apply plugin: 'com.neenbedankt.android-apt'
 apply plugin: 'me.tatarka.retrolambda'
 apply plugin: 'com.jakewharton.butterknife'
+apply plugin: 'com.squareup.sqldelight'
 
 android {
     compileSdkVersion rootProject.ext.androidCompileSdkVersion
@@ -12,6 +13,7 @@ android {
         minSdkVersion rootProject.ext.androidMinSdkVersion
         targetSdkVersion rootProject.ext.androidTargetSdkVersion
         multiDexEnabled true
+        vectorDrawables.useSupportLibrary = true
         versionCode 1
         versionName "1.0"
         buildConfigField "boolean", "CAN_JUMP", "true"
@@ -37,6 +39,7 @@ android {
         dev {
             buildConfigField "boolean", "CAN_JUMP", "true"
             applicationIdSuffix = ".dev"
+            buildConfigField "int", "EXOPACKAGE_FLAGS", "1"
         }
         prod {
             buildConfigField "boolean", "CAN_JUMP", "false"
@@ -45,6 +48,7 @@ android {
         demo {
             buildConfigField "boolean", "CAN_JUMP", "false"
             applicationIdSuffix = ".demo"
+            buildConfigField "int", "EXOPACKAGE_FLAGS", "1"
         }
     }
 
@@ -97,7 +101,14 @@ dependencies {
     compile "com.android.support:recyclerview-v7:${rootProject.ext.androidSupportSdkVersion}"
     compile "com.squareup.leakcanary:leakcanary-android:1.3.1"
 
-    apt 'com.google.dagger:dagger-compiler:2.0.1'
+    apt 'com.google.dagger:dagger-compiler:2.5'
+    apt "com.ryanharter.auto.value:auto-value-gson:0.3.1"
+    apt "com.google.auto.value:auto-value:1.2"
+    // used for AndroidStudio sync
+    provided "com.google.auto.value:auto-value:1.2"
+    // to make OkBuck work with SqlDelight
+    provided 'com.github.okbuilds:SqlDelightBin:0.4.3'
+
     devCompile project(path: ':dummylibrary', configuration: 'freeRelease')
     prodCompile project(path: ':dummylibrary', configuration: 'paidRelease')
     demoCompile project(path: ':dummylibrary', configuration: 'paidRelease')
--- a/build.gradle
+++ b/build.gradle
@@ -1,6 +1,12 @@
 buildscript {
+    configurations.all {
+        resolutionStrategy {
+            dependencySubstitution {
+                substitute module('com.tunnelvisionlabs:antlr4') with module('org.antlr:antlr4:4.5.3')
+            }
+        }
+    }
     repositories {
-        mavenLocal()
         jcenter()
     }
     dependencies {
@@ -8,12 +14,12 @@ buildscript {
         classpath 'com.neenbedankt.gradle.plugins:android-apt:1.8'
         classpath 'me.tatarka:gradle-retrolambda:3.2.5'
         classpath 'com.jakewharton:butterknife-gradle-plugin:8.2.1'
+        classpath 'com.squareup.sqldelight:gradle-plugin:0.4.3'
     }
 }
 
 allprojects {
     repositories {
-        mavenLocal()
         jcenter()
     }
 }
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -37,6 +37,14 @@ task sourcesJar(type: Jar) {
     classifier = 'sources'
 }
 
+configurations.all {
+    resolutionStrategy {
+        dependencySubstitution {
+            substitute module('com.tunnelvisionlabs:antlr4') with module('org.antlr:antlr4:4.5.3')
+        }
+    }
+}
+
 dependencies {
     compile gradleApi()
     compile localGroovy()
@@ -44,7 +52,7 @@ dependencies {
     compile 'commons-io:commons-io:2.4'
     compile 'commons-lang:commons-lang:2.6'
     compile 'org.apache.maven:maven-artifact:3.3.9'
-    compile 'com.android.tools.build:gradle:2.1.0'
+    compile 'com.android.tools.build:gradle:2.1.2'
 
     testCompile 'junit:junit:4.12'
 }
--- a/buildSrc/src/main/groovy/com/github/okbuilds/core/model/AndroidTarget.groovy
+++ b/buildSrc/src/main/groovy/com/github/okbuilds/core/model/AndroidTarget.groovy
@@ -7,14 +7,11 @@ import com.android.manifmerger.ManifestMerger2
 import com.android.manifmerger.MergingReport
 import com.android.utils.ILogger
 import com.github.okbuilds.core.util.FileUtil
-import com.github.okbuilds.okbuck.ExperimentalExtension
-import groovy.transform.Memoized
 import groovy.transform.ToString
 import org.apache.commons.codec.digest.DigestUtils
 import org.gradle.api.JavaVersion
 import org.gradle.api.Project
 import org.gradle.api.logging.Logger
-
 /**
  * An Android target
  */
@@ -27,6 +24,8 @@ abstract class AndroidTarget extends JavaLibTarget {
     final int minSdk
     final int targetSdk
     final boolean generateR2
+    final Scope sqldelight
+    final File sqldelightDir
 
     AndroidTarget(Project project, String name) {
         super(project, name)
@@ -47,20 +46,40 @@ abstract class AndroidTarget extends JavaLibTarget {
 
         // Butterknife support
         generateR2 = project.plugins.hasPlugin('com.jakewharton.butterknife')
+
+        sqldelightDir = new File(project.projectDir, "build/okbuck/sqldelight")
+        // SqlDelight support
+        if (project.plugins.hasPlugin('com.squareup.sqldelight')) {
+            sqldelightDir.mkdirs()
+            sqldelight = new Scope(project, ["provided"] as Set, [sqldelightDir] as Set)
+        } else {
+            sqldelight = null
+        }
     }
 
     protected abstract BaseVariant getBaseVariant()
 
     protected abstract ManifestMerger2.MergeType getMergeType()
 
+    String getSqlDelightJar() {
+        String runner = sqldelight.externalDeps.find { it.contains("SqlDelightBin") }
+        if (runner == null) {
+            throw new IllegalStateException("SqlDelight model module must declare the SqlDelightBin dependency, " +
+                    "read more at https://github.com/OkBuilds/OkBuck/wiki/Known-caveats#sqldelight-work-around")
+        }
+        return project.rootProject.projectDir.absolutePath + "/" + runner
+    }
+
     @Override
     Scope getMain() {
+        Set<File> srcDirs = baseVariant.sourceSets.collect { SourceProvider provider ->
+            provider.javaDirectories
+        }.flatten() as Set<File>
+
         return new Scope(
                 project,
                 ["compile", "${buildType}Compile", "${flavor}Compile", "${name}Compile"] as Set,
-                baseVariant.sourceSets.collect { SourceProvider provider ->
-                    provider.javaDirectories
-                }.flatten() as Set<File>,
+                srcDirs,
                 null,
                 extraJvmArgs)
     }
--- a/buildSrc/src/main/groovy/com/github/okbuilds/okbuck/generator/BuckFileGenerator.groovy
+++ b/buildSrc/src/main/groovy/com/github/okbuilds/okbuck/generator/BuckFileGenerator.groovy
@@ -149,6 +149,11 @@ final class BuckFileGenerator {
             ":${rule.name}"
         }
 
+        // SqlDelight
+        if (target.sqldelight != null) {
+            rules.add(GenSqlDelightRuleComposer.compose(target))
+        }
+
         // Lib
         androidLibRules.add(AndroidLibraryRuleComposer.compose(
                 target,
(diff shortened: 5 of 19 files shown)


build.gradle
```diff
@@ -4,20 +4,25 @@ buildscript {
...
+        classpath 'com.squareup.sqldelight:gradle-plugin:0.4.2'
     }
 }
 
 allprojects {
     repositories {
         mavenLocal()
         jcenter()
+        maven {
+            url  "http://dl.bintray.com/okbuild/maven"
```
kageiit: this should not be required right? everything should be on jcenter already?

Piasy: SqlDelightBin is not on jcenter yet, I've created a request on it.

buildSrc/src/main/groovy/com/github/okbuilds/okbuck/generator/BuckFileGenerator.groovy
```diff
@@ -133,6 +133,18 @@ final class BuckFileGenerator {
             ":${rule.name}"
         }
 
+        // SqlDelight
+        if (target.sqldelight != null) {
+            String runner = target.sqldelight.externalDeps.find { it.contains("SqlDelightBin") }
+            if (runner == null) {
+                throw new IllegalStateException("SqlDelight model module must declare the dependency: " +
+                        "`provided(group: 'com.github.okbuilds', name: 'SqlDelightBin', version: '1.0.0', ext: 'jar', classifier: '')`")
```
kageiit: we should leave version of SqlDelightBin out of the error.

Actually okbuck plugin can automatically apply this provided dependency if sqldelight plugin is enabled. so users dont need to specify explicitly unless they want to change version

buildSrc/src/main/groovy/com/github/okbuilds/okbuck/rule/GenSqlDelightRule.groovy
```diff
@@ -0,0 +1,30 @@
...
+        mSqlDelightRunnerPath = sqlDelightRunnerPath
+        mProjectDir = projectDir
+    }
+
+    @Override
+    final void print(PrintStream printer) {
+        printer.println("genrule(")
+        printer.println("\tname = '${name}',")
+        printer.println("\tsrcs = glob([")
+        printer.println("\t\t'src/main/sqldelight/**/*.sq',")
```
kageiit: why just src/main? can you just iterate over the srcDirs instead of this target?

buildSrc/src/main/groovy/com/github/okbuilds/okbuck/generator/BuckFileGenerator.groovy
Piasy: Great, I'll try to add it automatically.

buildSrc/src/main/groovy/com/github/okbuilds/okbuck/rule/GenSqlDelightRule.groovy
Piasy: sure

build.gradle
kageiit: Cool lets wait on that before merging if you dont mind

kageiit:
Just minor comments. Rest looks good

buildSrc/src/main/groovy/com/github/okbuilds/okbuck/generator/BuckFileGenerator.groovy
Piasy: Could we add dependency into the project after resolve?

If not, I think we should't add the SqlDelightBin dependency arbitrarily, it should only be added when user really use SqlDelight.

I'll change the error message to refer to our wiki page.

Piasy:
Updated

.travis.yml
```diff
@@ -11,10 +11,7 @@ android:
     - extra-android-support
 
 script:
-  - bash ./gradlew okbuck installBuck &&
```
kageiit: Can you please restore this back to how it was?

app/build.gradle
```diff
@@ -83,7 +87,14 @@ dependencies {
...
     compile "com.squareup.leakcanary:leakcanary-android:1.3.1"
 
-    apt 'com.google.dagger:dagger-compiler:2.0.1'
+    apt 'com.google.dagger:dagger-compiler:2.5'
+    apt "com.ryanharter.auto.value:auto-value-gson:0.3.1"
+    apt "com.google.auto.value:auto-value:1.2"
+    // used for AndroidStudio sync
+    provided "com.google.auto.value:auto-value:1.2"
+    // to make OkBuck work with SqlDelight
+    provided(group: 'com.github.okbuilds', name: 'SqlDelightBin', version: '1.0.0', ext: 'jar', classifier: '')
```
kageiit: This should be automatic right?

build.gradle
```diff
@@ -1,23 +1,33 @@
 buildscript {
+    configurations.all {
+        resolutionStrategy {
+            dependencySubstitution {
+                substitute module('com.tunnelvisionlabs:antlr4') with module('org.antlr:antlr4:4.5.3')
```
kageiit: why do we need this?

build.gradle
```diff
@@ -1,23 +1,33 @@
...
+        classpath 'com.squareup.sqldelight:gradle-plugin:0.4.2'
     }
 }
 
 allprojects {
     repositories {
-        mavenLocal()
         jcenter()
+        maven {
+            url  "http://dl.bintray.com/okbuild/maven"
```
kageiit: is this still needed?

buildSrc/build.gradle
```diff
@@ -37,14 +37,22 @@ task sourcesJar(type: Jar) {
     classifier = 'sources'
 }
 
+configurations.all {
+    resolutionStrategy {
+        dependencySubstitution {
+            substitute module('com.tunnelvisionlabs:antlr4') with module('org.antlr:antlr4:4.5.3')
```
kageiit: why do we need this?

buildSrc/src/main/groovy/com/github/okbuilds/core/model/JavaLibTarget.groovy
```diff
@@ -22,6 +23,14 @@ class JavaLibTarget extends JavaTarget {
         } else {
             retrolambda = null
         }
+
+        // SqlDelight
+        if (project.plugins.hasPlugin('com.squareup.sqldelight')) {
+            sqldelight = new Scope(project, ["provided"] as Set,
+                    [new File(project.projectDir, "build/generated/source/sqldelight")] as Set)
```
kageiit: build/generated/source/sqldelight does not exist at project configuration time right?

Also what about debug vs release variants for the generated sources?

.travis.yml
Piasy: sure, but travis build fail, really strange

app/build.gradle
Piasy: Could we add dependency into the project after resolve?

If not, I think we should't add the SqlDelightBin dependency arbitrarily, it should only be added when user really use SqlDelight.

build.gradle
Piasy: SqlDelight has problem work with android gradle plugin, see https://github.com/square/sqldelight/issues/123

build.gradle
Piasy: no need anymore, I'll remove it

buildSrc/src/main/groovy/com/github/okbuilds/core/model/JavaLibTarget.groovy
Piasy: SqlDelight does not support variants. 

Can we create it before create the sqldelight scope?

Piasy:
Travis fails because the source files list is generated during BUCK's configuration time (I guess), that means it's generated before my `genrule` runs, so BUCK cannot find the generated sources at its first compile round, although my `genrule` will generate Java files during this time. If I run buck build after its first fail, it will succeed. That's why it build successfully in my local machine.

@kageiit Do you have any idea how to fix this problem?

buildSrc/src/main/groovy/com/github/okbuilds/core/model/JavaLibTarget.groovy
```diff
@@ -22,6 +23,17 @@ class JavaLibTarget extends JavaTarget {
         } else {
             retrolambda = null
         }
+
+        // SqlDelight
+        if (project.plugins.hasPlugin('com.squareup.sqldelight')) {
+            File sqldelightDir = new File(project.projectDir, "build/generated/source/sqldelight")
```
kageiit: why is this the same as the one the gradle plugin generated sources end up in?

Can we not use a temp directory for okbuck purposes only?

buildSrc/src/main/groovy/com/github/okbuilds/core/model/JavaLibTarget.groovy
```diff
@@ -22,6 +23,17 @@ class JavaLibTarget extends JavaTarget {
...
+            if (!sqldelightDir.exists()) {
```
kageiit: dont need this check mkdirs exits gracefully

buildSrc/src/main/groovy/com/github/okbuilds/core/model/JavaLibTarget.groovy
```diff
@@ -22,6 +23,17 @@ class JavaLibTarget extends JavaTarget {
...
+                sqldelightDir.mkdirs()
+            }
+            sqldelight = new Scope(project, ["provided"] as Set, [sqldelightDir] as Set)
```
kageiit: this should be in android target if you want to specify provided

if it works for java libs as well, you need to support both provided and compileOnly configurations

buildSrc/src/main/groovy/com/github/okbuilds/okbuck/generator/BuckFileGenerator.groovy
```diff
@@ -133,6 +133,18 @@ final class BuckFileGenerator {
...
 
+        // SqlDelight
+        if (target.sqldelight != null) {
+            String runner = target.sqldelight.externalDeps.find { it.contains("SqlDelightBin") }
+            if (runner == null) {
+                throw new IllegalStateException("SqlDelight model module must declare the SqlDelightBin dependency, " +
+                        "read more at https://github.com/OkBuilds/OkBuck/wiki/Known-caveats#sqldelight-work-around")
+            }
+            rules.add(GenSqlDelightRuleComposer.compose(target,
+                    target.project.rootProject.projectDir.absolutePath + "/" + runner))
```
kageiit: This seems to reach in too deep to an implementation detail `target.project.rootProject.projectDir.absolutePath + "/" + runner`
We should expose the path to the sqldeligght jar via the target api similar to the retrolamdba jar

kageiit:
Added more comments inline.

I think to fix the travis issue, the real fix is to have the src_target.name rules depend on the sqldelight_target.name rules

buildSrc/src/main/groovy/com/github/okbuilds/core/model/JavaLibTarget.groovy
Piasy: Sure, no problem

buildSrc/src/main/groovy/com/github/okbuilds/core/model/JavaLibTarget.groovy
Piasy: SqlDelight only works with Android module, I'll move this to android target.

buildSrc/src/main/groovy/com/github/okbuilds/okbuck/generator/BuckFileGenerator.groovy
Piasy: Updated

buildSrc/src/main/groovy/com/github/okbuilds/core/model/AndroidTarget.groovy
```diff
@@ -46,12 +47,33 @@ abstract class AndroidTarget extends JavaLibTarget {
 
         // Butterknife support
         generateR2 = project.plugins.hasPlugin('com.jakewharton.butterknife')
+
+        // SqlDelight support
+        if (project.plugins.hasPlugin('com.squareup.sqldelight')) {
+            File sqldelightDir = new File(project.projectDir, "build/okbuck/sqldelight")
+            if (!sqldelightDir.exists()) {
```
kageiit: dont need to check if it exists

buildSrc/src/main/groovy/com/github/okbuilds/okbuck/generator/BuckFileGenerator.groovy
```diff
@@ -133,6 +133,12 @@ final class BuckFileGenerator {
             ":${rule.name}"
         }
 
+        // SqlDelight
+        if (target.sqldelight != null) {
+            rules.add(GenSqlDelightRuleComposer.compose(target))
+            deps.add(":sqldelight_${target.name}")
```
kageiit: i dont think this will work.

the sqldelight genrule needs to run before the src_target.name rule is built.

You may need to extract all sql files to a separate android module to make this work correctly

buildSrc/src/main/groovy/com/github/okbuilds/core/model/AndroidTarget.groovy
Piasy: // this dir is set as source set, if it doesn't exist, then we can't read it later

buildSrc/src/main/groovy/com/github/okbuilds/okbuck/generator/BuckFileGenerator.groovy
Piasy: understand, I'll try it

buildSrc/src/main/groovy/com/github/okbuilds/core/model/AndroidTarget.groovy
kageiit: i mean you can directly call `mkdirs()`

if it exists, it will continue quietly. if it does not, it will create it

no need for the if block

Piasy: Great, I'll remove the check.

kageiit:
Ill take a pass at this

Going to squash all the commits into one to make this a bit easier/cleaner

Piasy:
@kageiit With [fb guys' help in stackoverflow](http://stackoverflow.com/a/38808109/3077508), I finally make SqlDelight work with BUCK, but if I rebase this branch to latest master, two errors will be thrown:

``` bash
aidl E 53910 1563601 aidl.cpp:537] refusing to generate code from aidl file defining parcelable

BUILD FAILED: com.squareup.javapoet.CodeBlock.of(Ljava/lang/String;[Ljava/lang/Object;)Lcom/squareup/javapoet/CodeBlock;
```

By trying incremental rebase, I find out that the second error is introduced by 38269effa0a7343e4e029d0c1b1b7a265f3fbe16, and the first error is introduced by 3baaccf97b2abbf5da0baccf5e78a952ac758ec6.

build.gradle
kageiit: Seems like they fixed this issue. can we remove the workaround?

kageiit:
@Piasy can you please do a rebase and push so I can look at this better? It will be also easier to find a solution to the problemsyou posted then

Piasy:
No problem, and this pr passed ci yesterday night, but after I squash the commits, it fails.

I'll rebase it.

app/build.gradle
```diff
@@ -97,7 +101,14 @@ dependencies {
...
     compile "com.squareup.leakcanary:leakcanary-android:1.3.1"
 
-    apt 'com.google.dagger:dagger-compiler:2.0.1'
+    apt 'com.google.dagger:dagger-compiler:2.5'
+    apt "com.ryanharter.auto.value:auto-value-gson:0.3.1"
+    apt "com.google.auto.value:auto-value:1.2"
+    // used for AndroidStudio sync
+    provided "com.google.auto.value:auto-value:1.2"
+    // to make OkBuck work with SqlDelight
+    provided 'com.github.okbuilds:SqlDelightBin:0.4.3'
```
kageiit: do we have to maintain this everytime a new version of sqldelight is released?

This does not seem ideal. We should open a PR on sqldelight so they can upload another artifact that does what we want so their changes will keep the binary compatible

Piasy: I suggested, but Jake said they don't want to include this in :(

kageiit: Hmm that sucks. Can we actually create a jar that can put the gradle plugin on the classpath and just call the public api to do what it needs to. That way we can switch out the sqldelight plugin jar with the version from gradle and still have the same result. Lesser maintanence for us too. Similar to how we do retrolambda

Piasy: The SQL to Java compilation should happen during buck build, not okbuck task, so I create a executable jar to call SqlDelight's internal api. I think its public api is only the gradle task, but we can't use that.

kageiit: Can you push the sqldelightbin code under okbuck in github so others can contribute if a newer version is released and they need to use it?

I worry about this because the sqldelight plugin version might not match what users use in gradle. We should put a warning message to the output when this happens so they are aware

Piasy: Source code is already in github: https://github.com/OkBuilds/SqlDelightBin.

kageiit: Hold off on pushing the jar. going to explore the alternative approach we discussed offline

kageiit:
Merged in #160

PR #131, opened 2016-07-11
Remove ':' from alias names

### LINKED ISSUE #130: Buck aliases generated from Gradle modules in directories are not valid alias names.
If you have a Gradle project with many modules - and some modules are in a folder like this:
```
* module
* moduleGroup
  * appModule
```

OkBuck will generate an alias named `moduleGroup:appModule` - this is not a valid buck alias.

kageiit:
Fixes #130


tonycosentini:
👍 thanks!

PR #134, opened 2016-07-23
Support to add extra buck options to rules

kageiit:
Add an option to enable extra options on `android_binary` rules. These rules by far have the most number of undocumented options which makes it hard to support them all officially without adding bloat. We can later extend this to provide support for other rules as needed.

Going to leave this undocumented in readme as this api is still in flux.


raviagarwal7:
looks good!

PR #136, opened 2016-07-26
Extract target dependencies and populate cache before creating rule to make sure all projects are on the same version of dependencies.

raviagarwal7:
- Buck file generator goes through each project one by one and to gets all its rules.
 - If a rule is created for a project which is very low in (near leaf) in the hierarchy then the resolved dependencies (in the rule) might differ from its parent ones (or from the project in a completely disjoint tree)
 - This diff is resolving configuration and extracting dependencies before creating rules.

### DIFF
--- a/buildSrc/src/main/groovy/com/github/okbuilds/core/model/Target.groovy
+++ b/buildSrc/src/main/groovy/com/github/okbuilds/core/model/Target.groovy
@@ -56,4 +56,8 @@ abstract class Target {
     def getProp(Map map, defaultValue) {
         return map.get("${identifier}${name}", map.get(identifier, defaultValue))
     }
+
+    public void resolve() {
+        // no scope to resolve
+    }
 }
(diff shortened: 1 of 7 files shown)


buildSrc/src/main/groovy/com/github/okbuilds/core/model/Target.groovy
```diff
@@ -56,4 +56,8 @@ abstract class Target {
     def getProp(Map map, defaultValue) {
         return map.get("${identifier}${name}", map.get(identifier, defaultValue))
     }
+
+    public void resolve() {
```
kageiit: Make this abstract?

buildSrc/src/main/groovy/com/github/okbuilds/okbuck/generator/BuckFileGenerator.groovy
```diff
@@ -60,8 +72,6 @@ final class BuckFileGenerator {
             }
         }
 
-        createProjectConfigRule(project, project.okbuck.projectTargets)
```
kageiit: Why is this removed

raviagarwal7: we are not using the return value here, it's already being called above - https://github.com/OkBuilds/OkBuck/pull/136/files#diff-8e0e8d98076e9ca8aa632421f239491eL30

PR #141, opened 2016-07-29
Improve okbuckClean task

kageiit:
- Remove all generated `BUCK` files (not just those in subprojects). This is useful when projects move to different directories.
- Ignore generated `BUCK` files in `.okbuck`
- Do not remove `.buckd` and`buck-out`directories by default


kageiit:
@Piasy just a heads up, changing the default behavior of the okbuckClean task to be buck cache friendly

cc @nguyentruongtho

kageiit:
Also added the first tests for the plugin!

tonycosentini:
🚢

Issue #145, opened 2016-08-05
Http Server port not set error when used with the Buck IntelliJ plugin.


ragunathjawahar:
````
com.facebook.buck.util.HumanReadableException: Your buck server may be turned off, since the Buck daemon is on port -1.
Try adding to your '.buck.local' or '.buckconfig' file, if you don't have it already set:
[httpserver]
    port = 0
After that, restart IntelliJ or reopen your project.
````

kageiit:
This is intentionally not set because some may want to use a real http server. Suggestion is to set this in `.buckconfig` instead of `.buckconfig.local` and check the file into vcs

PR #146, opened 2016-08-05
Fix for issue #145.

ragunathjawahar:


kageiit:
This is intentionally not set because some may want to use a real http server. Suggestion is to set this in .buckconfig instead of .buckconfig.local and check the file into vcs

PR #149, opened 2016-08-08
Add support to fecth source jars

kageiit:
- Use group name in cached file names to avoid potential conflicts
- Fetch source jars from gradle cache if available


Piasy:
I just wonder why do we need fetch source jars?

kageiit:
> I just wonder why do we need fetch source jars?

We are not downloading them separately. We are just linking against them if they are available so when we do `buck project`, we get source navigation in the IDE when that project is imported

PR #159, opened 2016-08-21
Generate java_binary Rule

Leland-Takamine:
Generate java_binary rule for projects that use ApplicationPlugin.


kageiit:
Looks good mostly. Two comments

There appears to be an unused main constant in Java binary rule

Please add an integration test by adding a sample java application project in examples

Thanks for contributing!

kageiit:
nvm, the constant appears to be used. I am going to merge this for now and look at adding tests in a later commit

PR #160, opened 2016-08-23
SqlDelight Support

kageiit:
- Provides a generic framework to hook any third party gradle plugins that generate sources inside buck
- Add sqldelight support as an example

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,15 +9,22 @@ android:
     - tools
     - platform-tools
     - android-24
-    - build-tools-24.0.1
+    - build-tools-24.0.2
     - extra-android-m2repository
     - extra-android-support
+  licenses:
+    - android-sdk-preview-license-84831b94
+    - android-sdk-license-2be07077
+    - android-sdk-license-.+
+    - google-gdk-license-.+
 
 script:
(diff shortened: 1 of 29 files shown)


kageiit:
build error seems unrelated. will fix it in a bit. but verified this works locally.

kageiit:
This seems to run fine locally, but travis does not like multiple gradle processes invoked by buck.

@Piasy can you test this out on your machine?

Piasy:
I've tested it yesterday, it works fine in my machine too.

kageiit:
ok im trying some solutions from https://discuss.gradle.org/t/timeout-waiting-to-lock-artifact-cache/2545/14

kageiit:
> ok im trying some solutions from https://discuss.gradle.org/t/timeout-waiting-to-lock-artifact-cache/2545/14

This did not work, so i put in a workaround to not have the buildSrc folder lying around when running inside buck (since it is not needed)

Piasy:
Hope to see a release including this :)

kageiit:
I am trying to get in espresso test support
plan to cut a release with that as well

Piasy:
Great! I can wait for that 👍

PR #167, opened 2016-09-15
Move BuildConfig application ID to base AndroidTarget

seanabraham:
This moves the ApplicationId part of the build config generator to the base AndroidTarget in a similar fashion to: https://github.com/OkBuilds/OkBuck/commit/3e3d818fe6ced9d15ea6962f54b73841fe12c1ca

This will generate the following for a library:

```python
android_build_config(
    ...
	package = 'com.package.lib',
	values = [
         ...
		'String APPLICATION_ID = "com.package.lib"',
         ...
	],
    ...
)
```

And remain the same for an application:

```python
android_build_config(
    ...
	package = 'com.package.app',
	values = [
         ...
		'String APPLICATION_ID = "com.package.app"',
         ...
	],
    ...
)
```

### DIFF
--- a/buildSrc/src/main/groovy/com/github/okbuilds/core/model/AndroidTarget.groovy
+++ b/buildSrc/src/main/groovy/com/github/okbuilds/core/model/AndroidTarget.groovy
@@ -132,6 +132,7 @@ abstract class AndroidTarget extends JavaLibTarget {
 
     List<String> getBuildConfigFields() {
         List<String> buildConfig = [
+            "String APPLICATION_ID = \"${applicationId + applicationIdSuffix}\"",
             "String BUILD_TYPE = \"${buildType}\"",
             "String FLAVOR = \"${flavor}\"",
         ]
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/groovy/com/github/okbuilds/core/model/AndroidTarget.groovy
```diff
@@ -145,6 +145,7 @@ abstract class AndroidTarget extends JavaLibTarget {
             String key, ClassField classField ->
                 "${classField.type} ${key} = ${classField.value}"
         }
+        buildConfig.add("String APPLICATION_ID = \"${applicationId + applicationIdSuffix}\"")
```
kageiit: Can you put this at top, so we have alpha ordering of the generated fields

seanabraham:
Okay, moved it to the top @kageiit. PTAL

buildSrc/src/main/groovy/com/github/okbuilds/core/model/AndroidTarget.groovy
seanabraham: 👍 , done.

PR #185, opened 2016-09-29
Add support for external arguments when running okbuck via buckw

seanabraham:
Sometimes there's a need for additional args to be passed to gradle (i.e. on CI) which wind up resolving the project differently which in turn affect the buck configuration. This change allows end users to set the `OKBUCK_ARGS` env variable (and renames `EXTRA_ARGS` -> `CLEAN_ARGS` for clarity) so that any additional parameters can be passed as necessary. Open to naming suggestions here.

### DIFF
--- a/buckw
+++ b/buckw
@@ -128,11 +128,11 @@ runOkBuck ( ) {
     if [[ ! -z "$INSTALLED_WATCHMAN" ]]; then
         getToClean | jsonq '"\n".join(obj["files"])' | xargs rm
         info "DELETED OLD BUCK FILES"
-        CLEAN_ARGS="-xokbuckClean"
+        CLEAN_ARG="-xokbuckClean"
     fi
 
     rm -f $OKBUCK_SUCCESS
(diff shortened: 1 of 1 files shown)


kageiit:
You can just check for the System property

`okbuck.wrapper` and see if it is true to change your configuration in your gradle scripts.

seanabraham:
Per offline discussion, we only want to pass flags in some cases (e.g. on CI). Also renaming to `OKBUCK_EXTRA_ARGS`.

buckw
```diff
@@ -128,11 +128,11 @@ runOkBuck ( ) {
     if [[ ! -z "$INSTALLED_WATCHMAN" ]]; then
         getToClean | jsonq '"\n".join(obj["files"])' | xargs rm
         info "DELETED OLD BUCK FILES"
-        EXTRA_ARGS="-xokbuckClean"
+        CLEAN_ARG="-xokbuckClean"
```
kageiit: Can we call this `OKBUCK_ARGS`

We have some changes coming for incremental okbuck and the name would be more appropriate

kageiit: Actually, this can be cleaned up later as well. I will merge this for now

seanabraham: 👍

PR #190, opened 2016-10-03
Adds support for annotationProcessor dependencies

felipecsl:


kageiit:
Please add some tests by using this configuration in one of the example projects

felipecsl:
@kageiit is this what you had in mind? I confirmed manually that the dependencies were correctly added to `provided_deps` on the BUCK file

kageiit:
> @kageiit is this what you had in mind? I confirmed manually that the dependencies were correctly added to provided_deps on the BUCK file

If travis is green, everything is good. Thanks for checking manually as well :)

PR #201, opened 2016-10-08
Use package instead of applicationId inside resource rule

marcinkwiatkowski:
The `package` element inside main `AndroidManifest.xml` specifies the package `R.java`, the `applicationId` may be different for different flavors, but `package` is always the same.

I'm happy to change the implementation, wanted to get something working first.

### DIFF
--- a/buildSrc/src/main/groovy/com/uber/okbuck/core/model/AndroidTarget.groovy
+++ b/buildSrc/src/main/groovy/com/uber/okbuck/core/model/AndroidTarget.groovy
@@ -11,7 +11,6 @@ import com.android.manifmerger.ManifestMerger2
 import com.android.manifmerger.MergingReport
 import com.android.utils.ILogger
 import com.uber.okbuck.core.util.FileUtil
-import groovy.transform.Memoized
 import groovy.transform.ToString
 import groovy.util.slurpersupport.GPathResult
 import groovy.xml.StreamingMarkupBuilder
@@ -36,6 +35,8 @@ abstract class AndroidTarget extends JavaLibTarget {
     final int targetSdk
     final boolean debuggable
     final boolean generateR2
+    String manifest
+    String packageName
 
     AndroidTarget(Project project, String name) {
         super(project, name)
@@ -68,6 +69,8 @@ abstract class AndroidTarget extends JavaLibTarget {
             minSdk = baseVariant.mergedFlavor.minSdkVersion.apiLevel
             targetSdk = baseVariant.mergedFlavor.targetSdkVersion.apiLevel
         }
+
+        ensureManifest()
     }
 
     protected abstract BaseVariant getBaseVariant()
@@ -222,7 +225,7 @@ abstract class AndroidTarget extends JavaLibTarget {
         }.flatten() as Set<String>
     }
 
-    String getManifest() {
+    private void ensureManifest() {
         Set<String> manifests = [] as Set
 
         baseVariant.sourceSets.each { SourceProvider provider ->
@@ -230,7 +233,7 @@ abstract class AndroidTarget extends JavaLibTarget {
         }
 
         if (manifests.empty) {
-            return null
+            return
         }
 
         File mainManifest = project.file(manifests[0])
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/groovy/com/uber/okbuck/core/model/AndroidTarget.groovy
```diff
@@ -221,13 +221,19 @@ abstract class AndroidTarget extends JavaLibTarget {
...
-        baseVariant.sourceSets.each { SourceProvider provider ->
-            manifests.addAll(getAvailable(Collections.singletonList(provider.manifestFile)))
+    String getPackage(){
+        Set manifests = getManifestsFromSourceSets()
+        if (manifests.empty) {
+            return null
         }
 
+        XmlSlurper slurper = new XmlSlurper()
+        GPathResult manifestXml = slurper.parse(project.file(manifests[0]))
```
kageiit: what if secondary manifests define this attribute instead?

buildSrc/src/main/groovy/com/uber/okbuck/core/model/AndroidTarget.groovy
```diff
@@ -221,13 +221,19 @@ abstract class AndroidTarget extends JavaLibTarget {
         }.flatten() as Set<String>
     }
 
-    String getManifest() {
-        Set<String> manifests = [] as Set
-
-        baseVariant.sourceSets.each { SourceProvider provider ->
-            manifests.addAll(getAvailable(Collections.singletonList(provider.manifestFile)))
+    String getPackage(){
```
kageiit: This can be a potentially expensive operation to parse manifest for every `android_resource` in a target

We should add `@Memoized` to this as you mentioned this does not change

kageiit:
I think it's cleaner if we left getManifest alone and just store the packageId during the parsing done in getManifest. Which can then be returned via getPackage

marcinkwiatkowski:
Yea, I thought about it, I guess the drawback is that `getManifest` would have to be called before `getPackage` and that depends on the order of rule composers.

kageiit:
You can extract the manifest extraction to a private method and call it something like ensureManifest and just store it in a field and use it for whichever method gets called first

kageiit:
@marcinkwiatkowski Have you signed the [CLA](https://docs.google.com/a/uber.com/forms/d/1pAwS_-dA1KhPlfxzYLBqK6rsSWwRwH95OCCZrcsY5rk/viewform) ?

buildSrc/src/main/groovy/com/uber/okbuck/core/model/AndroidTarget.groovy
```diff
@@ -257,7 +261,8 @@ abstract class AndroidTarget extends JavaLibTarget {
             XmlSlurper slurper = new XmlSlurper()
             GPathResult manifestXml = slurper.parse(project.file(mergedManifest))
 
-            manifestXml.@package = applicationId + applicationIdSuffix
+            packageName = manifestXml.@package
```
kageiit: I dont think this is correct.

See http://tools.android.com/tech-docs/new-build-system/applicationid-vs-packagename

kageiit: If you look at the intermediate manifest gradle generates, it has applicationId + suffix in it

Like
```
<manifest xmlns:android="http://schemas.android.com/apk/res/android"
    package="com.uber.okbuck.example.demo.debug"
```

kageiit:
Looks like the package name from the main manifest ((if set, otherwise the applicationId from build.gradle) decides the package for the generated R file.

The final package name in the merged manifest file is still applicationId + applicationIdSuffix (if set, otherwise it falls back to the value specified in the base manifest)

We need to preserve both the characteristics

marcinkwiatkowski:
Yes, signed the CLA and fixed the package name in merged manifest. 
Looks like the package element always is required in main manifest and the merger fails if it's missing. `applicationId` on the other hand is optional is defaults to package specified in the main manifest.

PR #208, opened 2016-10-11
Don't merge manifest in the constructor

marcinkwiatkowski:
`AndroidTarget` is also used for external dependencies.


buildSrc/src/main/groovy/com/uber/okbuck/core/model/AndroidTarget.groovy
```diff
@@ -225,6 +224,20 @@ abstract class AndroidTarget extends JavaLibTarget {
         }.flatten() as Set<String>
     }
 
+    String getPackage() {
+        if (packageName == null) {
```
kageiit: Juat do if(!packageName)

PR #211, opened 2016-10-11
Fixed when initial boot class path is empty or null

malbano:


buildSrc/src/main/groovy/com/uber/okbuck/core/model/JavaLibTarget.groovy
```diff
@@ -60,11 +60,14 @@ class JavaLibTarget extends JavaTarget {
     }
 
     String getBootClasspath() {
-        String bootCp = initialBootCp
+        List<String> classpaths = []
+        if (initialBootCp != null && !initialBootCp.isEmpty()) {
```
kageiit: if(initialBootCp)

checks both

PR #213, opened 2016-10-12
Fixed find dependencies command to exclude dummy and abi

malbano:
Fixed find dependencies command to exclude dummy and abi jar files.

### DIFF
--- a/buildSrc/src/main/groovy/com/uber/okbuck/printable/PostProcessClassessCommands.groovy
+++ b/buildSrc/src/main/groovy/com/uber/okbuck/printable/PostProcessClassessCommands.groovy
@@ -21,7 +21,13 @@ public class PostProcessClassessCommands implements Printable {
     @Override
     public void print(PrintStream printer) {
         if (!mPostprocessClassesCommands.isEmpty()) {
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/groovy/com/uber/okbuck/printable/PostProcessClassessCommands.groovy
```diff
@@ -21,7 +21,13 @@ public class PostProcessClassessCommands implements Printable {
     @Override
     public void print(PrintStream printer) {
         if (!mPostprocessClassesCommands.isEmpty()) {
-            String deps = "\$(JARS=(`find ${mGenDir} ! -name \"*-abi.jar\" ! -name \"*dex.dex.jar\" -name \"*.jar\"`); " +
+            List<String> depsFindConstraints = []
+            depsFindConstraints.add("! -name \"*-abi.jar\"")
+            depsFindConstraints.add("! -path \"*#dummy*/abi.jar\"")
+            depsFindConstraints.add("! -name \"*#dummy*.jar\"")
```
kageiit: this should cover the above line already

malbano: It doesn't. The first excludes paths containing `*#dummy*` with filename `abi.jar`, while the second excludes filenames with `*#dummy*.jar`

PR #214, opened 2016-10-12
Fix buck wrapper CUSTOM_BUCK_REPO on OSX

felipecsl:
the `md5digest` function was adding leading blank lines and spaces to the hash
on OSX, causing the `CUSTOM_BUCK_REPO` override to not work, since 
`git remote add` would reject a name with newlines.

Paired on it with @seanabraham


kageiit:
We actually dont need to use md5digest anymore after #212
But seems like @raviagarwal7 only fixed the wrapper and not the template. He will open a PR to fix the template as well

PR #227, opened 2016-11-03
Robolectric fails to load resources from depending modules

felipecsl:
Still not sure whether Robolectric or Buck issue or if there is a
workaround we could do in the test runner, but putting this up here for
now just for documentation/discussion/feedback.
Added a failing test that demostrates the issue.

### DIFF
--- a/libraries/common/src/test/java/com/uber/okbuck/example/common/ExampleUnitTest.java
+++ b/libraries/common/src/test/java/com/uber/okbuck/example/common/ExampleUnitTest.java
@@ -2,13 +2,12 @@ package com.uber.okbuck.example.common;
 
 import android.content.Context;
 
-import org.junit.Test;
-import org.junit.runner.RunWith;
-
 import com.uber.okbuck.BuckRobolectricTestRunner;
 
-import org.robolectric.annotation.Config;
+import org.junit.Test;
+import org.junit.runner.RunWith;
 import org.robolectric.RuntimeEnvironment;
+import org.robolectric.annotation.Config;
 
 import static org.junit.Assert.assertEquals;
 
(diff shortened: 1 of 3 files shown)


libraries/common/src/test/java/com/uber/okbuck/example/common/ExampleUnitTest.java
```diff
@@ -25,5 +24,8 @@ public void addition_isCorrect() throws Exception {
     public void resource_loading() {
         Context context = (Context) RuntimeEnvironment.application;
         assertEquals(context.getResources().getString(R.string.app_name), "Common");
+        assertEquals(context.getResources().getColor(
+            com.uber.okbuck.example.parcelable.R.color.fooColor), 0xcccccc);
```
kageiit: This is probably not the right example. This specific problem goes away when you use the Right R class here.

Can you modify this change to repro your case i.e have another module refer to the resource in its main code and this module invokes that piece of code in it's test?

felipecsl: Yes, you're right. Good point.

libraries/parcelable/src/main/java/com/uber/okbuck/example/common/Person.java
```diff
@@ -1,8 +1,11 @@
 package com.uber.okbuck.example.common;
 
+import android.content.Context;
 import android.os.Parcel;
 import android.os.Parcelable;
 
+import com.uber.okbuck.example.parcelable.R;
```
kageiit: does this error happen if you use `com.uber.okbuck.example.common.R` ?

felipecsl: @kageiit test updated. `com.uber.okbuck.example.common.R` works, but `com.uber.okbuck.example.parcelable.R` does not

felipecsl: That's the one I'm using, `com.uber.okbuck.example.parcelable.R` is the merged R for this library

kageiit:
@seanabraham @felipecsl This seems to be working if you use the default robolectric test runner that comes as part of robolectric (which handles the buck build system correctly now). I rebased your branch to the latest okbuck master on the felipecsl-robolectric-resources branch. Feel free to check the same and  I can close this ticker

PR #228, opened 2016-11-06
Android Lint Support

kageiit:
- Adds support for generating `lint_*` `genrule`s for `android_library` targets
- Hooks are in place to let lint rules be generated for any jvm library targets. Theoretically lint can be run on java, kotlin or scala etc with some small tweaks. Lint can also be run on test sources with a few small tweaks as well. These changes will be part of a subsequent PR
- Supports custom lint jars embedded in aars
- Supports custom lint project/external dependencies. For example, a java library that defines custom lint rules can be added to another library as a linter by making it a dependency of the `buckLint` configuration
- Fully incremental, parallel and buck cache/network cache friendly
- Supports many android dsl options specified in `lintOptions`
- Ability to pin lint api version and not be surprised when android gradle plugin pulls in new checks on updates
- Requires an updated `buckw` by running the `buckWrapper` to watch for changes to `lint.xml` files and run `okbuck` again as needed
- Lint on all android rules can be run like `buck query "filter('lint_*', kind('genrule', '//...'))" | xargs buck build`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,7 @@ script:
   - mv buildSrc notBuildSrc
   - buck query "kind('android_binary', '//...')" | xargs buck build
   - buck test --include unit
+  - buck query "filter('lint_*', kind('genrule', '//...'))" | xargs buck build
   - android-wait-for-emulator
   - adb shell input keyevent 82
   - buck test //app:instrumentation_demoDebug_test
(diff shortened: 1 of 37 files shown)


kageiit:
I will add some examples for all the new lint features and make changes to travis as well in a follow up commit

PR #234, opened 2016-11-09
Use file path to resolve dependency name

marcinkwiatkowski:


buildSrc/src/main/groovy/com/uber/okbuck/core/model/Scope.groovy
```diff
@@ -87,9 +94,10 @@ class Scope {
                 configuration.files.findAll { File resolved ->
                     !resolvedFiles.contains(resolved)
                 }.each { File localDep ->
+                    String localDepPath = rootPath.relativize(Paths.get(localDep.absolutePath)).toString()
```
kageiit: You can just use Fileutil.relativePath from okbuck here instead of bringing in all new imports. we can refactor that util method later to use the nio apis.

using nio makes this plugin only work if you have jdk 1.8

buildSrc/src/main/groovy/com/uber/okbuck/core/model/Scope.groovy
```diff
@@ -34,6 +40,7 @@ class Scope {
           DependencyCache depCache = OkBuckGradlePlugin.depCache) {
 
         this.project = project
+        this.rootPath = Paths.get(project.gradle.rootProject.projectDir.absolutePath)
```
kageiit: why is this a field? We can just do `project.rootDir` to get the root project dir inline

Issue #232, opened 2016-11-08
Bazel support


aj-michael:
Hey guys, I saw on your [Wiki](https://github.com/uber/okbuck/wiki/OkBuck-Overview) a comment about extending okbuck to support other build systems like Bazel. I was wondering what your plans are for supporting Bazel in particular and wanted to discuss the possibility of contributing a pull request to extend the okbuck plugin to generate Bazel BUILD files.

I have a rough first draft of the implementation [here](https://github.com/aj-michael/okbuck/commit/203c5f684416cd52909b369c9dc21cef2d3bf883). It's not as fully featured as okbuck and it could use some polish before I send a pull request. I tried to keep all of the code self contained in it's own package and not touch any okbuck internals.

I think okbuck is a great project for helping people transition off of Gradle or even to just experiment with new things. Is adding Bazel support something you'd be interested in?

kageiit:
Hey @aj-michael Thanks for making the first implementation of okbazel. We have always wanted to start working on bazel support but we also wanted to make sure all the tools from gradle can be ported and work seamlessly on an alternative build systme. We finally added lint support recently and are working on adding transform support next.

Part of the reason we did not move forward with bazel originally was the lack of test rules for android and project generation/IDE support. buck also had all the rules built for ios and it is a huge advantage to have both platforms use the same build system at scale.

We are definitely interested in adding bazel support, but we want to focus our energies on getting everything to work on one build system which supports the most features and start looking at the other options shortly after.

I can take the okbazel PR in an experimental module, but we will not be able to maintain it if something else in the project breaks. This will change as we finish up all the android gradle plugin features in buck and look at alternate build system support. Please open the PR after cleaning up and adding as much documentation as you can.

Thanks a lot for pushing this project forward!

aj-michael:
Hey @kageiit , that's fantastic! Sorry for the delayed response, the last week has been very busy for me.

As an update: I am currently working on polishing it up and fixing a bug I discovered due to differences in resource merging between Bazel and Buck for AARs. I plan to send the pull request near the end of this week.

A few questions first:

1. By experimental module, do you just mean that all documentation will indicate that it is experimental, or do you mean the package name contains "experimental" in it? At the moment, my code is confined to the "com.uber.okbuck.bazel" package. Is there a better place that you would prefer it live?
[...]

PR #251, opened 2016-11-21
Experimental okbazel support

aj-michael:
See #232 for context

Documentation for this change is in the commit description of a5a7f9e and in a separate [buildSrc/src/main/groovy/com/uber/okbuck/experimental/bazel/README.md](https://github.com/uber/okbuck/pull/251/files#diff-4f4f9322475c6d0f4ab2081a290a12c2)

Sorry for the delay on sending this, I had a bit of difficultly getting the cache to work correctly due to some differences between how Buck and Bazel handle resource dependencies.


aj-michael:
Hmmm, I looked through the Travis CI failure and it doesn't appear to be related to this set of commits.

hzsweers:
Thanks! FYI - @kageiit is in Hawaii this week on vacation, so probably won't get reviewed until next week.

aj-michael:
Ok, thanks for the heads up 😃

kageiit:
Ill take a look by end of this week. Can you check the travis failure?

aj-michael:
Sorry for the delayed response, I've been busy this week. I've looked at the Travis failure and it appears unrelated to this change? I'm not sure how to re-trigger Travis to run to check if it is just a flake. As far as I know, my change should not impact any tests as they don't use the OkBazel plugin.

kageiit:
I restarted the build. But for future reference, you can close and reopen a PR to have travis rekick your build

buildSrc/src/main/groovy/com/uber/okbuck/experimental/bazel/AndroidBinaryRule.groovy
```diff
@@ -0,0 +1,42 @@
+package com.uber.okbuck.experimental.bazel
+
+import com.uber.okbuck.rule.BuckRule
+
+final class AndroidBinaryRule extends BuckRule {
```
kageiit: Can we name all these classes with a `Bazel` prefix? This is to avoid confusing with the buck classes

buildSrc/src/main/groovy/com/uber/okbuck/experimental/bazel/AndroidLibraryRule.groovy
```diff
@@ -0,0 +1,66 @@
...
+final class AndroidLibraryRule extends BuckRule {
+    private static final String RULE_TYPE = "android_library"
+
+    private final Set<String> srcSet
+    private final String manifest
+    private final Set<String> providedDeps
+    private final String packageName
+    private final Set<AndroidTarget.ResBundle> resources
+
+    AndroidLibraryRule(
```
kageiit: does bazel not support annotation processors currently? or is it done implicitly somehow?

buildSrc/src/main/groovy/com/uber/okbuck/experimental/bazel/AndroidLibraryRuleComposer.groovy
```diff
@@ -0,0 +1,28 @@
+package com.uber.okbuck.experimental.bazel
+
+import com.uber.okbuck.composer.AndroidBuckRuleComposer
+import com.uber.okbuck.core.model.AndroidLibTarget
+
+final class AndroidLibraryRuleComposer extends AndroidBuckRuleComposer {
```
kageiit: I would suggest extending the `AndroidBuckRuleComposer` to create a `BazelAndroidBuckRuleComposer` which can then override the definitions of `external` etc. to keep the logic cleaner

buildSrc/src/main/groovy/com/uber/okbuck/experimental/bazel/OkBazelGradlePlugin.groovy
```diff
@@ -0,0 +1,86 @@
...
+                    workspaceFile.write WORKSPACE
+                }
+                OkBuckGradlePlugin.depCache =
+                        new DependencyCache(project, CACHE_PATH, true, null, false, true)
+
+                BuildFileGenerator.generate(project).each { subProject, buildFile ->
+                    PrintStream printer = new PrintStream(subProject.file("BUILD")) {
+                        @Override
+                        public void println(String s) {
+                            // BUILD files are typically space-delimited. Since
```
kageiit: Great point. I think we should allow customizing the delimiters easily. But this works for now

buildSrc/src/main/groovy/com/uber/okbuck/experimental/bazel/README.md
```diff
@@ -0,0 +1,77 @@
...
+    ./gradlew okbazel
+
+This will generate one `BUILD` file for each Gradle subproject, one `BUILD` file
+in `okbazel/BUILD` and one `WORKSPACE` file. The `WORKSPACE` file contains an
+`android_sdk_repository` rule which will contain the path of your Android SDK if
+the `$ANDROID_HOME` environment variable is set. You will need to update
+`WORKSPACE` to contain the `api_level` and `build_tools_version` from your SDK
+that you wish to use.
+
+## Setting up Bazel
```
kageiit: I would recommend investing in a `bazelw` script that will download and install bazel from source similar to what `buckw` does. It can also run okbuck when files have changed similar to `buckw`. I think this is a very necessary usability feature for adoption.

buildSrc/src/main/groovy/com/uber/okbuck/experimental/bazel/OkBazelGradlePlugin.groovy
aj-michael: The Bazel team maintains [buildifier](https://github.com/bazelbuild/buildifier), a tool for formatting BUILD files according to the standard convention. I added this hack so that the BUILD files generated by okbuck adhere to that convention. As an alternative, we could invoke buildifier to clean up the BUILD files at the end. Although, the integration may be messy as buildifier is written in go.

kageiit: I meant customizing delimiters in okbuck, not bazel specifically, but that works as well :)

buildSrc/src/main/groovy/com/uber/okbuck/experimental/bazel/AndroidBinaryRule.groovy
aj-michael: Done.

buildSrc/src/main/groovy/com/uber/okbuck/experimental/bazel/AndroidLibraryRuleComposer.groovy
aj-michael: Ah, I just wrote out a long reply about this and then I think github went down for 2 minutes.

The gist of it was, I've added a commit that introduces BazelRuleComposer. It's a trait as opposed to a subclass of BuckRuleComposer because I need to extend both AndroidBuckRuleComposer and JavaBuckRuleComposer, which each extend BuckRuleComposer.

aj-michael:
I'm still working on addressing all of the comments so far, it's just a busy week for me. I will get to all of them. Are you planning to add more comments than the ones so far?

buildSrc/src/main/groovy/com/uber/okbuck/experimental/bazel/AndroidLibraryRule.groovy
aj-michael: Ok, for documentations sake, the way that annotation processors work in Bazel is we create a [`java_plugin`](https://bazel.build/versions/master/docs/be/java.html#java_plugin) rule that depends on a `java_library` that builds that annotation processor or a `java_import` that wraps a prebuilt JAR. The `java_plugin` specifies a `processor_class`. Then we put the `java_plugin` in the deps of an `android_library` or `java_library` that it should run on.

buildSrc/src/main/groovy/com/uber/okbuck/experimental/bazel/README.md
aj-michael: I will look into how `buckw` works and respond back on this thread. However, I don't think installing bazel from source is a possibility. Bazel can only be built with a pre-existing installation of Bazel. We do this to maintain platform independence (e.g. we don't want to be shipping pre-compiled, platform-specific binaries in the Bazel source tree that would be needed to bootstrap from scratch).

Furthermore, Bazel is very carefully versioned and, since it is pre-1.0, occasionally makes breaking changes. As such, we have some users who intentionally do not update to the newest version immediately because it may require a change to their `BUILD` files. We keep a history of platform-specific installers [here](https://github.com/bazelbuild/bazel/releases) and a detailed changelog for each release [here](https://github.com/bazelbuild/bazel/blob/master/CHANGELOG.md).

I agree that re-running okbuck when files have changed is important for usability.

kageiit: we should atleast have a filed in `bazelw` that will be able to download a particular version of bazel that is prebuilt if it is not buildable from source

aj-michael: Ok, I've added a `bazelWrapper` Gradle task that generates `bazelw`. `bazelw` downloads and executes the installer for bazel 0.4.2 (the most recent stable release at the time of writing). However, it is currently hardcoded to the Linux installer.

aj-michael:
Ok, I think I have now made it through all of the open comments.

buildSrc/src/main/resources/com/uber/okbuck/core/util/wrapper/BAZELW_TEMPLATE
```diff
@@ -0,0 +1,187 @@
...
+        # Install Bazel
+        wget -qO bazel_installer.sh https://github.com/bazelbuild/bazel/releases/download/0.4.2/bazel-0.4.2-installer-linux-x86_64.sh
+        bash bazel_installer.sh --user
+        BAZEL_BINARY=$HOME/bin/bazel
+      fi
+    fi
+}
+
+# Run tasks before bazel command
+setupBuckRun ( ) {
```
kageiit: Seems like a lot of common logic is reused across both wrappers. We use a common shell script internally for various wrapper scripts. Will push that change up so this script can get quite small

aj-michael: Yes, the main difference between these scripts is how Buck/Bazel is installed.

kageiit:
Can you perform a quick rebase? We made a ton of changes over the past week and want to ensure everything is in working order still

aj-michael:
Ok, working on the rebase now. As expected there are some cleanups that I will need to make.

aj-michael:
Ok, rebase is complete. I fixed some import paths and confirmed that it still works using `bazelw` on https://github.com/gradle/perf-android-large.

aj-michael:
Hey, really sorry about the delay on this. I took a long holiday, and have been catching up on other work in the few days I've been back. I've just finished the rebase and testing to ensure that everything is still good. PTAL :smile:

kageiit:
Can you add 

`apply plugin: 'com.uber.okbazel'` to okbuck build.gradle itself.

There are some issues with being able to use both plugins because the bazel plugin is trying to create the same extension as the buck plugin

aj-michael:
I asked about sharing the plugin id back in https://github.com/uber/okbuck/issues/232#issuecomment-260689659 , and yes you are correct, this prevents you from using both plugins in the same project. At the time, the reason I did that was because there were several places where the plugin id was used in the codebase and was hardcoded to okbuck.

I can take another look to see what it would take to give it a separate plugin ID.

kageiit:
Closing due to inactivity. Feel free to comment anytime if you would like me to re-open this

steeve:
@aj-michael did you ever found a way around this? great work though :)

PR #267, opened 2016-12-06
Add option to generate BUCK files in parallel

kageiit:
Adds an experimental `parallel` flag that lets okbuck generate BUCK files as a per project task. This can reduce okbuck run times when used in combination with gradle's parallel flag.


kageiit:
Flaky ui tests

PR #289, opened 2016-12-15
Add genrule support for running infer.

clonetwin26:
Only missing thing here is to pass the boot classpath to infer so it can correctly analyze the base classes like java.concurrent.


kageiit:
This can be done purely via a `DEFS` file as the rules are all generic enough to be composable directly by buck instead. See example in https://github.com/facebook/infer/pull/558

PR #310, opened 2017-01-04
Example of custom Application class not being created when using buck

chrislacy:
Run another-app via a gradle build. Observe the “D/MainActivity:
MainActivity.onCreate()
application:com.uber.okbuck.example.anotherapp.MainApplication” log
message.

When running with buck (via ./buckw install --run another-appDebug),
the log message is “D/MainActivity: MainActivity.onCreate()
application:android.app.Application”.


chrislacy:
Note: commenting out the exopackage in the root build.gradle file does result in the custom MainApplication class being used.

PR #312, opened 2017-01-05
Fix NDK environment variable name

chrislacy:


kageiit:
buck needs `ANDROID_NDK` to be set. Where did you get the idea for `ANDROID_NDK_HOME`?

chrislacy:
From the following log when running `./gradlew :buckWrapper`:

```
NDK is missing a "platforms" directory.
If you are using NDK, verify the ndk.dir is set to a valid NDK directory.  It is currently set to /usr/local/opt/android-sdk/ndk-bundle.
If you are not using NDK, unset the NDK variable from ANDROID_NDK_HOME or local.properties to remove this warning.
```

I just figured it was a typo. Feel free to ignore.

kageiit:
I would recommend installing via

`brew install android-ndk` and setting

`export ANDROID_NDK='brew prefix android-ndk'`

instead

Ignore the `ANDROID_NDK_HOME` part.

PR #316, opened 2017-01-07
Add extension to specify lint libraries.

clonetwin26:
This allows you to modify what gets put on the libraries flag via gradle dsl. You can use the buckLintLibrary configuration to put a dependency on the linter --libraries.

```
dependencies {
     buckLintLibrary deps.uber.javaRx
 }
```

The above would make a libraries flag that would look like the following.

```
--libraries .okbuck/cache/com.ubercab.rx.java-0.11.1.jar:.okbuck/cache/com.ubercab.rx.irrelevant-0.11.1.jar:.okbuck/cache/com.android.support.support-annotations-25.0.1.jar:.okbuck/cache/io.reactivex.rxjava-1.2.5.jar 
```

### DIFF
--- a/buildSrc/src/main/groovy/com/uber/okbuck/composer/android/LintRuleComposer.groovy
+++ b/buildSrc/src/main/groovy/com/uber/okbuck/composer/android/LintRuleComposer.groovy
@@ -56,6 +56,14 @@ final class LintRuleComposer extends JvmBuckRuleComposer {
 
         if (!target.main.sources.empty) {
             lintCmds.add("--classpath ${toLocation(":${src(target)}")}")
+
+            Set<String> lintLibraries = []
+            lintLibraries = lintLibraries + target.lintLibraries.externalDeps
+            lintLibraries = lintLibraries + target.lintLibraries.targetDeps
+            if (!lintLibraries.isEmpty()) {
+                String libraries = lintLibraries.join(':')
+                lintCmds.add("--libraries ${libraries}")
+            }
         }
         if (target.lintOptions.abortOnError) {
             lintCmds.add("--exitcode")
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/groovy/com/uber/okbuck/composer/android/LintRuleComposer.groovy
```diff
@@ -56,6 +56,13 @@ final class LintRuleComposer extends JavaBuckRuleComposer {
 
         if (!target.main.sources.empty) {
             lintCmds.add("--classpath ${toLocation(":${src(target)}")}")
+            if (!lintExtension.libraries.isEmpty()) {
+                String libraries = ""
+                lintExtension.libraries.each { library ->
```
hzsweers: nit: guava can do a joiner on `:` for you to one line this

clonetwin26: yes.

hzsweers:
Can you add an example snippet of how to configure this in the okbuck extensions?

clonetwin26:
added above.

hzsweers:
:+1:

kageiit:
@clonetwin26 the configuration should be done on a per project basis. We cannot apply a global config for all consumers. Its better to add a configuration to every buck project like `buckLintLibrary` and then every project can have their own deps

hzsweers:
would that be per lint project or per-every-project-that-wants-to-be-linted?

clonetwin26:
This still needs some work on getting what is put on --libraries sanitized

hzsweers:
Any word on what the scope of that will be?

kageiit:
Its most of the way there. I can pick up the last commit to join the dots

kageiit:
Being handled by #352

PR #325, opened 2017-01-13
Fix memory leak when using the gradle daemon

kageiit:
Move all statics globals to be instance variables on the plugin object instead


buildSrc/src/main/groovy/com/uber/okbuck/core/util/LintUtil.groovy
```diff
@@ -79,13 +77,14 @@ class LintUtil {
     }
 
     static DependencyCache getLintDepsCache(Project project) {
-        if (!lintDepCache) {
-            lintDepCache = new DependencyCache("lint",
+        OkBuckGradlePlugin okBuckGradlePlugin = project.rootProject.plugins.getPlugin(OkBuckGradlePlugin)
```
raviagarwal7: probably add a util method in projectutil to get okBUckGradlePlugin

raviagarwal7:
looks good!

PR #339, opened 2017-01-19
Fix regression with target deps not being picked up from buckLint configurations

kageiit:


kageiit:
cc @raviagarwal7 for future reference. Any Scope that needs to know about project/target deps needs to use a dependency cache instance that has `okbuck.buckProjects` made available when it is created

PR #352, opened 2017-01-25
Add ability to add lint libraries to genrules

kageiit:
This is useful when using newer PSI based lints which can enhance the AST with proper type information when used with the `--libraries` flag.

To make the classes of any external or project dependency available to the buck linter for analysis, they can be specified as

```
dependencies {
  buckLintLibray "com.example:dep:1.0.0"
  buckLintLibray project(":dep")
}
```

### DIFF
--- a/buildSrc/src/main/groovy/com/uber/okbuck/OkBuckGradlePlugin.groovy
+++ b/buildSrc/src/main/groovy/com/uber/okbuck/OkBuckGradlePlugin.groovy
@@ -46,6 +46,7 @@ class OkBuckGradlePlugin implements Plugin<Project> {
     public static final String DEFAULT_CACHE_PATH = ".okbuck/cache"
     public static final String GROUP = "okbuck"
     public static final String BUCK_LINT = "buckLint"
+    public static final String BUCK_LINT_LIBRARY = "buckLintLibrary"
     public static final String LINT = "lint"
     public static final String TRANSFORM = "transform"
     public static final String RETROLAMBDA = "retrolambda"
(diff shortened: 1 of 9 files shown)


kageiit:
Checked that a sample generated `BUCK` file had a lint genrule with the `--libraries` flag pointing to the right dependencies.

hzsweers:
Could you add an integration test of sorts with one of our RxJava lints?

kageiit:
> Could you add an integration test of sorts with one of our RxJava lints?

@clonetwin26 added an integration test for lint in 8c53f195912e59938495fc2d4d4588fd2c6e51e6

We can enhance it by adding a similar rule that takes advantage of the PSI based lints

kageiit:
> We can enhance it by adding a similar rule that takes advantage of the PSI based lints

@hzsweers to follow up with adding a new lint rule to take advantage of the new capabilities

felipecsl:
heads up that the summary of this PR has a typo `buckLintLibray` instead of `buckLintLibrary` 😉  careful when copy-pasting!

felipecsl:
I've tried adding intellij psi as a buck lint library dependency but still it wont run any of the PSI checks. Any other gotchas?

```
    buckLintLibrary 'com.android.tools.external.lombok:lombok-ast:0.2.3'
    buckLintLibrary 'com.android.tools.external.com-intellij:uast:145.597.3'
    buckLintLibrary 'org.ow2.asm:asm:5.0.4'
    buckLintLibrary 'org.ow2.asm:asm-tree:5.0.4'
```

hzsweers:
@felipecsl the intention is to put libraries that your lint engine needs to run against on the classpath like this, not necessarily core lint utilities themselves. @kageiit correct me if I'm wrong

kageiit:
That is correct

PR #358, opened 2017-01-27
Test case for lint PSI checks not running with Buck

felipecsl:
This PR is a test case for demonstrating that Buck doesn't seem to correctly run PSI based lint checks.

How to verify:

```
./gradlew libraries:lintErrorLibrary:lint # 1 error found
./buckw build //libraries/lintErrorLibrary:lint_release # no errors found
```

### DIFF
--- a/dependencies.gradle
+++ b/dependencies.gradle
@@ -44,7 +44,7 @@ def external = [
 
 def lint = [
         lint      : "com.android.tools.lint:lint:25.1.0",
-        lintApi   : "com.android.tools.lint:lint-api:25.1.0",
+        lintApi   : "com.android.tools.lint:lint-api:25.2.2",
         lintChecks: "com.android.tools.lint:lint-checks:25.1.0",
         lintTests : "com.android.tools.lint:lint-tests:25.1.0",
 ]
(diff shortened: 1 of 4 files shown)


kageiit:
With the test case, even `./gradlew libraries:lintErrorLibrary:lint` is not exiting with any errors. Can you make sure the check actually fails in gradle so we can compare with the buck linter?

kageiit:
btw you do not need to make any changes whatsoever to `libraries/customLintLibrary/build.gradle` and `libraries/lintErrorLibrary/build.gradle` from master.

clonetwin26:
Maybe we have to bump the build tools version to get latest lint?
https://github.com/uber/okbuck/blob/master/dependencies.gradle#L20

felipecsl:
hmm looks like `compile project (':libraries:customLintLibrary')` is not enough to have gradle pick up the custom lint check. If I use the install task then it works fine.

felipecsl:
Seems like `lintChecks` instead of `compile` works

kageiit:
@felipecsl Can you please try to rebase this PR on master? UAST/Psi lints should be picking up things correctly now

kageiit:
Assuming this works correctly and closing due to inactivity. Feel free to comment if there is an issue and we can definitely followup

PR #361, opened 2017-01-28
Add ability to reference resources between main and flavor res directories

### LINKED ISSUE #221: Support for extra resource directories
We don't define flavors, they just exist and Gradle is happy. They once existed as flavors but are no longer defined that way. Probably due to some magic under the hood. Either way, here is the situation:

I have the folder `module/main/res` with an `android_resource` configuration named `res_main`. There is a folder `module/x/res` that is also defined. This is the inexplicit flavor. OkBuck sees it fine, it defines a BUCK configuration for it named `res_x`. In `module/x/res` we reference a string in the main resources. Buck isn’t able to find that string because `res_x`isn’t declaring `res_main` as a dependency.

I've found this edge case for our project but could be likely others have seen it. Simply adding `res_main` as a dependency of `res_x` solved the issue.

I've started investigating myself but I find myself low on time. Filing this issue to perhaps encourage others to help or to just make others aware of this. Looks like OkBuck just simply needs to properly define the resource dependency.

kageiit:
Fixes #221
This is finally possible due to an update to buck in https://github.com/facebook/buck/commit/3319cd6ce5c0126fea859932cc5c2494e95b63b9

You need to use this sha or later in `.buckversion` to get support for this feature and continue using okbuck. This also greatly simplifies the number of `android_resource` rules and now okbuck only creates one per target regardless of the number of its `res` and `asset` directories.


kageiit:
cc @cwoodwar6 @Yuliya-Kaleda @MichaelEvans

Issue #375, opened 2017-01-31
OkBuck generates incorrect Buck file with test artifact


cwoodwar6:
Dependencies.gradle:
```
ext.deps = [
    dagger2: [
        dagger2: "com.google.dagger:dagger:2.8",
        dagger2compiler: "com.google.dagger:dagger-compiler:2.8",
        dagger2Test: "com.google.dagger:dagger:2.8:tests"
    ]
]
```

App module dependencies:
```
  compile project(':app-library')

  compile deps.dagger2.dagger2
  annotationProcessor deps.dagger2.dagger2compiler
  testCompile deps.dagger2.dagger2Test
```
Library module dependencies:
```
  compile deps.dagger2.dagger2
  annotationProcessor deps.dagger2.dagger2compiler
```

The library module's generated Buck file:
```
deps = [
   '//.okbuck/cache:com.google.dagger.dagger-2.8-tests.jar',
]
```
Dagger's test artifact is used in the library module instead of the main dagger artifact.

This line:
```//.okbuck/cache:com.google.dagger.dagger-2.8-tests.jar```
should instead be:
```//.okbuck/cache:com.google.dagger.dagger-2.8.jar```

Repro project: https://github.com/cwoodwar6/okbuck-test-project

kageiit:
I think this is because the dependency cache does not handle artifacts with classifiers correctly yet.

cc @raviagarwal7 

We will work on a fix, but in the meanwhile, you can try to rename your artifact to something like `com.google.dagger:dagger-tests:2.8` and upload to a local maven. The other option is to consume the dagger test jar as a local jar dependency to unblock yourself.

PR #382, opened 2017-02-03
Issue #375 - Add jar classifiers to the dependency cache.

cwoodwar6:


cwoodwar6:
Blocked on https://github.com/cla-assistant/cla-assistant/issues/198

buildSrc/src/main/groovy/com/uber/okbuck/core/dependency/ExternalDependency.groovy
```diff
@@ -23,6 +23,10 @@ class ExternalDependency extends VersionlessDependency {
         }
 
         this.depFile = depFile
+
+        if (artifactType == "jar") {
```
kageiit: why do we need to filter here?

kageiit: I think the check should be if `artifactClassifier` exists, assign it

buildSrc/src/main/groovy/com/uber/okbuck/core/dependency/VersionlessDependency.groovy
```diff
@@ -10,13 +10,18 @@ class VersionlessDependency {
     final String group
     final String name
 
+    String classifier
```
kageiit: Make this final and pass in default null

buildSrc/src/main/groovy/com/uber/okbuck/core/dependency/ExternalDependency.groovy
cwoodwar6: When I was testing this, it included AARs. Those AARs had classifiers such as "debug" which seemed to break dependency resolution. This was just to keep it simple and not have to make any additional changes.

buildSrc/src/main/groovy/com/uber/okbuck/core/dependency/VersionlessDependency.groovy
cwoodwar6: Done. This change adds the additional arguments to the constructor and pushes off the type logic here as well.

buildSrc/src/main/groovy/com/uber/okbuck/core/dependency/VersionlessDependency.groovy
```diff
@@ -9,14 +9,21 @@ class VersionlessDependency {
 
     final String group
     final String name
+    final String classifier
 
-    VersionlessDependency(ModuleVersionIdentifier identifier) {
+    VersionlessDependency(ModuleVersionIdentifier identifier, String artifactType, String artifactClassifier) {
         group = identifier.group
         name = identifier.name
+        if (artifactType == "jar") {
```
kageiit: This is q very custom hack that only works for jar artifacts. I do not like how this is currently setup. aars can have classifiers as well when they are external dependencies.

I think the proper fix is to ignore the artifactType and classifier if it is not an external dependency.

kageiit:
Superceded by @388

Issue #400, opened 2017-02-22
Kotlin Support Tracker


kageiit:
- [x] Pure kotlin modules
- [x] Mixed kotlin/java modules
- [x] Android Kotlin modules
- [x] Kapt
- [x] Kotlin rule support for `buck project` https://github.com/facebook/buck/pull/1200

runningcode:
We've been slacking on keeping these github issues up to date.

Buck itself supports pure kotlin modules, mixed kotlin/java modules and android kotlin modules.
Buck is missing support for [kapt](https://github.com/facebook/buck/issues/956) and [robolectric_rule](https://github.com/facebook/buck/issues/1349). We should add robolectric_rule support to this list.

Also, okbuck support for mixed kotlin/java and android kotlin modules is on this branch: https://github.com/runningcode/okbuck/tree/no/kotlin-android-mixed.
They won't be upstreamed because @kageiit is planning on adding support in a large refactor of okbuck.

kageiit:
[Okbuck v0.22.0](https://github.com/uber/okbuck/releases/tag/v0.22.0) is out with Kotlin Android support and mixed Kotlin/Java modules support

kageiit:
@runningcode I already added `robolectric_test` support to okbuck. As soon as buck starts supporting it, it should start working out of the box

artem-zinnatullin:
What is the state of `kapt` support? Thanks!

kageiit:
It is blocked by https://github.com/facebook/buck/issues/956

talenguyen:
What about the status of this. I saw kotlin is supported in the [Features](https://github.com/uber/okbuck/wiki/Features) page.

kageiit:
Yes everything is supported except `kapt`. @cwoodwar6 is currently working on kapt support

erawhctim:
Forgive my ignorance, but what does partial kotlin support entail exactly? What does the current "partial" kotlin support (e.g. everything but `kapt`) get us? Can you build everything but annotation-processor-based things?

kageiit:
> Forgive my ignorance, but what does partial kotlin support entail exactly? What does the current "partial" kotlin support (e.g. everything but kapt) get us? Can you build everything but annotation-processor-based things?

Yes. If your gradle modules do not use the `kapt` configuration for annotation processors, they should work fine.

kassim:
It's looking hopeful https://github.com/cwoodwar6/buck/pull/3

aphexcx:
[...]

Issue #323, opened 2017-01-12
Kotlin support


vRallev:
I read that Kotlin is supported by Buck now. In the Fragment podcast it was mentioned that the Kotlin support is WIP. The [features wiki](https://github.com/uber/okbuck/wiki/Features) states that it's planned. But I couldn't find a ticket to track progress. Is there maybe an experimental snapshot build already?

kageiit:
There is no snapshot yet. Thanks for opening the issue. We are waiting on better kotlin support in buck (its quite basic right now) and plan to open the corresponding support PR in okbuck right after

kageiit:
Specifically waiting for https://github.com/brettwooldridge/buck/tree/kotlin-compiler to merge back into usptream. cc @brettwooldridge

brettwooldridge:
My current estimate is ~2 weeks before a pull request can be submitted...

vRallev:
Thanks, that sounds great. I'm also glad to hear that there's already a lot progress.

traviswestbrook:
Just wanted to give this a bump and see if there was anything that could be done to help out :)

kageiit:
Nothing much needed from the okbuck side. compiling mixed sources and annotation processor support is currently missing in buck. Creating the bridge between gradle to buck via okbuck once those features are available will not be much work

kageiit:
> Just wanted to give this a bump and see if there was anything that could be done to help out :)

You can help with the conversation in
https://github.com/facebook/buck/issues/1037
https://github.com/facebook/buck/issues/956

and see if you can help out with the buck codebase

seanabraham:
@brettwooldridge any updates on your end? Is this the latest on in-memory support? https://github.com/brettwooldridge/buck/commit/61fddf0ead19695b5d0ddafec96c34d98b698621

kageiit:
> My current estimate is ~2 weeks before a pull request can be submitted...

@brettwooldridge Any updates on the kotlin PR? Please let us know if there is any way we can move forward the kotlin in memory compiler support at the minimum

runningcode:
We should close this issue in favor of issue #400 .
In memory compiler support has been merged.
Buck now supports kotlin, kotlin android module, and kotlin and java mixed sources.

kageiit:
Duplicate of #400

PR #401, opened 2017-02-22
Preliminary Kotlin Support

kageiit:
Ref #400

This adds basic support only for building and testing pure kotlin modules. `kotlinc` is invoked out of memory and needs to be improved on buck's side to make it faster in the future. See #323

Generated kotlin rules look like

```
kotlin_library(
	name = 'src_main',
	srcs = glob([
		'src/main/kotlin/**/*.kt',
	]),
	tests = [
		':test_main',
	],
	source = '7',
	target = '7',
	deps = [
		'//.okbuck/cache:org.jetbrains.kotlin.kotlin-runtime-1.0.6.jar',
		'//.okbuck/cache:org.jetbrains.kotlin.kotlin-stdlib-1.0.6.jar',
	],
	visibility = [
		'PUBLIC',
	],
)

kotlin_test(
	name = 'test_main',
	srcs = glob([
		'src/test/kotlin/**/*.kt',
	]),
	source = '7',
	target = '7',
	labels = [
		'unit',
		'kotlin',
	],
	deps = [
		'//.okbuck/cache:junit.junit-4.12.jar',
		'//.okbuck/cache:org.hamcrest.hamcrest-core-1.3.jar',
		'//.okbuck/cache:org.jetbrains.kotlin.kotlin-runtime-1.0.6.jar',
		'//.okbuck/cache:org.jetbrains.kotlin.kotlin-stdlib-1.0.6.jar',
		'//.okbuck/cache:org.jetbrains.kotlin.kotlin-test-1.0.6.jar',
		'//.okbuck/cache:org.jetbrains.kotlin.kotlin-test-junit-1.0.6.jar',
		':src_main',
	],
	visibility = [
		'PUBLIC',
	],
)
```


buildSrc/src/main/java/com/uber/okbuck/core/util/KotlinUtil.java
```diff
@@ -0,0 +1,75 @@
...
+                ImmutableMap.of("template-kotlin-version", kotlinVersion));
+        kotlinc.setExecutable(true);
+
+        String kotlinCompiler = rootProject.relativePath(kotlinc);
+        String kotlinRuntime = String.format("%s/lib/%s-%s.jar", KOTLIN_HOME_LOCATION, KOTLIN_RUNTIME_MODULE,
+                kotlinVersion);
+        return Pair.of(kotlinCompiler, kotlinRuntime);
+    }
+
+    private static boolean findKotlin(ResolvedArtifact artifact) {
```
raviagarwal7: nit -- probably inline this

raviagarwal7:
Looks good! 
please add the generated kotlin buck rules for the fixture library in the summary for reference.

runningcode:
By "pure kotlin modules" do you mean kotlin modules without annotation processors or kotlin modules which do not include java source files?

kageiit:
> By "pure kotlin modules" do you mean kotlin modules without annotation processors or kotlin modules which do not include java source files?

See #400 

basically support now is only for a module that has `src/main/kotlin`. If you have java sources, you can make a java module and have your kotlin module depend on that (or vice versa). Mixed java/kotlin compilation is not supported by buck yet

kuno:
I am  playing okbuck 0.19.3 with kotlin 1.1.1 on an experimental project, which only contains an app module written in pure kotlin.

The module itself is only have a custom `Application` class and a default activity.

I can build it, installed on devices. But when running, it crashed caused by not found the application class and the default activity class in any path.

Is this the normal behavior at this age ?

PR #408, opened 2017-03-04
Keep cache of generated BUCK files path to improve buck file deletion.

raviagarwal7:
- Keep project path relative to root project in `.okbuck/state/STATE`
- Deletes the buck files which are not generated in the current run but generated in the last run.
- Saves the relative path of projects whose BUCK files are generated.
```
╰─$ cat .okbuck/state/STATE
another-app
app
dummylibrary
libraries
libraries/common
...
libraries/lintErrorLibrary
libraries/parcelable
libraries/robolectric-base
```
- Note: manually deleting the .okbuck/state/STATE file may lead failures with buck running on stale BUCK files as those won't be deleted on okbuck run.


raviagarwal7:
I think we should not have a fallback mechanism to delete using file tree api if `.okbuck/state/STATE `file is not present. 

If we have it then clean builds will try to delete all the BUCK files even though there is none. We will have to pay a one time cost of manually deleting the BUCK files from CI boxes (that too if it the upgrade coincides with someone removing a project)

buckw
```diff
@@ -129,20 +106,8 @@ runOkBuck ( ) {
...
-            EXTRA_OKBUCK_ARGS="-xokbuckClean $EXTRA_OKBUCK_ARGS"
-        else
-            warn "WATCHMAN FAILED. USING OKBUCKCLEAN."
-            removeWatch
-        fi
-    fi
-
     rm -f $OKBUCK_SUCCESS
-    ( $WORKING_DIR/gradlew -p $WORKING_DIR okbuck -Dokbuck.wrapper=true $EXTRA_OKBUCK_ARGS &&
+    ( $WORKING_DIR/gradlew -p $WORKING_DIR okbuck -Dokbuck.wrapper=true &&
```
kageiit: We still need EXTRA_OKBUCK_ARGS for other things. Please keep that

buildSrc/src/main/groovy/com/uber/okbuck/OkBuckGradlePlugin.groovy
```diff
@@ -37,6 +37,17 @@ import org.gradle.api.artifacts.repositories.FlatDirectoryArtifactRepository
...
 
+// Dependency Tree
+//
+//                okbuck
+//          /                \
+//         /             okbuckClean
+//        /        /          |          \
+//       /     :p1:okbuck   :p2:okbuck   :p3:okbuck     ...
+//      |        /           /               /
+//      setupOkbuck
```
kageiit: This does not look right `setupOkbuck` should run before all other `okbuck` tasks

raviagarwal7: will update it with arrows, task depends on the ones below.

buckw
raviagarwal7: will do

buckw
```diff
@@ -129,17 +106,7 @@ runOkBuck ( ) {
...
-        if [[ $? -eq 0 ]]; then
-            echo $TO_CLEAN | xargs rm
-            info "DELETED OLD BUCK FILES"
-            EXTRA_OKBUCK_ARGS="-xokbuckClean $EXTRA_OKBUCK_ARGS"
-        else
-            warn "WATCHMAN FAILED. USING OKBUCKCLEAN."
-            removeWatch
-        fi
-    fi
+    EXTRA_OKBUCK_ARGS=""
```
kageiit: revert this one line

buildSrc/src/main/groovy/com/uber/okbuck/OkBuckGradlePlugin.groovy
```diff
@@ -182,8 +185,17 @@ class OkBuckGradlePlugin implements Plugin<Project> {
                         kotlinDeps ? kotlinDeps.right: null)
             }
 
+            // Create clean task
+            Task okBuckClean = project.tasks.create(OKBUCK_CLEAN, OkBuckCleanTask, {
+                projects = okbuckExt.buckProjects
+            })
+            okBuckClean.setGroup(GROUP)
+            okBuckClean.setDescription("Delete stale configuration files generated by OkBuck")
+
```
kageiit: remove newline

buildSrc/src/main/java/com/uber/okbuck/core/task/OkBuckCleanTask.java
```diff
@@ -1,43 +1,73 @@
...
+                projects.stream()
+                        .map(project -> rootProjectPath
+                                .relativize(project.getProjectDir().toPath()).toString()
+                        )
+                        .collect(Collectors.toSet());
+
+        // Delete stale project's BUCK file
+        Sets.difference(lastProjectPaths, currentProjectPaths)
+                .stream()
+                .map(p -> rootProjectPath.resolve(p).resolve("BUCK"))
```
kageiit: ake `BUCK` a constant

PR #416, opened 2017-03-29
Fix .okbuck folder generation

seanabraham:
This PR does a few things:

1) Some of the files/subdirectories of the `.okbuck` directory were being created relative to where `gradlew` was being invoked from, as opposed to the root project dir. This fixes that so that everything under the `.okbuck` directory is created in the right spot.

2) Updates buckw so that `.okbuck` is looked for at `$WORKING_DIR/.okbuck` so that the check for previous successful okbuck runs works appropriately

3) Adds support to buck wrapper generation to ignore specific directories. We were doing this internally and I thought it would be helpful to upstream. This will make sure the watchman queries correctly not attempt to track changes under the `.okbuck` directory.

Our `buckw` internally also looks a bit different because our watchman project root is a folder above our android directory. This can be generically supported by looking at the output of `watchman watch-project` to check the root it's watching and whether there is a relative root specified. Those values can be saved and used for the queries. E.g. if `watchman watch-project` gives something like the following:

```
{
    "version": "4.7.0",
    "watch": "/Users/sean_abraham/projects",
    "watcher": "fsevents"
}
```

We can structure the watchman queries like:
```
PROJECT_ROOT="/Users/sean_abraham/projects"
watchman --output-encoding=json --no-pretty -j 2>&1 <<-EOT
["query", "$PROJECT_ROOT", {
    "since": "n:okbuck_trig",
    ...
}]
EOT
```

And if `watchman watch-project` gives something like this:
```
{
    "version": "4.7.0",
    "watch": "/Users/sean_abraham/project",
    "watcher": "fsevents",
    "relative_path": "android"
}
```

We can structure the watchman queries like:
```
PROJECT_ROOT="/Users/sean_abraham/projects"
RELATIVE_ROOT="android"
watchman --output-encoding=json --no-pretty -j 2>&1 <<-EOT
["query", "$PROJECT_ROOT", {
    "since": "n:okbuck_trig",
    "relative_root": "$RELATIVE_ROOT",
    ...
}]
EOT
```

This shouldn't be too hard to do but pretty annoying in bash so I decided against doing it in this PR.

@kageiit

### DIFF
--- a/buckw
+++ b/buckw
@@ -4,7 +4,7 @@
 ##
 ##  Buck wrapper script to invoke okbuck when needed, before running buck
 ##
-##  Created by OkBuck Gradle Plugin on : Mon Mar 06 13:56:01 PST 2017
+##  Created by OkBuck Gradle Plugin on : Wed Mar 29 15:03:12 PDT 2017
 ##
 #########################################################################
 
@@ -55,7 +55,7 @@ DEFAULT_BUCK_INSTALL_DIR="$HOME/.gradle/caches/okbuck/buck"
 CUSTOM_BUCK_REPO=""
 CUSTOM_REMOTE_NAME="custom"
 OKBUCK_SUCCESS="$WORKING_DIR/build/okbuck.success"
-OKBUCK_DIR=".okbuck"
+OKBUCK_DIR="$WORKING_DIR/.okbuck"
 MAX_DISPLAY_CHANGES=10
 
 ensureWatch ( ) {
@@ -75,13 +75,16 @@ getChanges ( ) {
             ["suffix", "gradle"],
             ["name", ["gradle-wrapper.properties", "lint.xml"]],
             ["match", "**/src/**/AndroidManifest.xml", "wholename"]
+        ],
+        ["not",
+            ["dirname", ".okbuck"]
         ]
     ],
     "fields": ["name"]
 }]
 EOT`
     SOURCE_ROOTS=`watchman --output-encoding=json --no-pretty -j 2>&1 <<-EOT
-["query", ".", {
+["query", "$WORKING_DIR", {
     "since": "n:okbuck_source_roots",
     "expression": ["allof",
         ["type", "d"],
@@ -89,6 +92,9 @@ EOT`
             ["match", "**/src/**/java", "wholename"],
             ["match", "**/src/**/res", "wholename"],
             ["match", "**/src/**/resources", "wholename"]
+        ],
+        ["not",
+            ["dirname", ".okbuck"]
         ]
     ],
     "fields": ["new", "exists", "name"]
--- a/buckw
+++ b/buckw
@@ -4,7 +4,7 @@
 ##
 ##  Buck wrapper script to invoke okbuck when needed, before running buck
 ##
-##  Created by OkBuck Gradle Plugin on : Wed Mar 29 15:03:12 PDT 2017
+##  Created by OkBuck Gradle Plugin on : Wed Mar 29 16:37:26 PDT 2017
 ##
 #########################################################################
 
@@ -159,7 +159,7 @@ setupBuckBinary ( ) {
             if [[ -z "$REMOTE_EXISTS" ]]; then
                 git remote add $CUSTOM_REMOTE_NAME $CUSTOM_BUCK_REPO
             fi
-            cd $WORKING_DIR
+            cd -
         fi
 
         BUCK_BINARY="$BUCK_HOME/bin/buck"
--- a/buildSrc/src/main/groovy/com/uber/okbuck/composer/android/KeystoreRuleComposer.groovy
+++ b/buildSrc/src/main/groovy/com/uber/okbuck/composer/android/KeystoreRuleComposer.groovy
@@ -15,7 +15,7 @@ final class KeystoreRuleComposer {
     static String compose(AndroidAppTarget target) {
         AndroidAppTarget.Keystore keystore = target.keystore
         if (keystore != null) {
-            File storeDir = target.project.rootProject.file(".okbuck/keystore/${target.identifier.replaceAll(':', '_')}")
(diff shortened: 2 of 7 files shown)


buildSrc/src/main/groovy/com/uber/okbuck/composer/android/KeystoreRuleComposer.groovy
```diff
@@ -15,7 +15,7 @@ final class KeystoreRuleComposer {
     static String compose(AndroidAppTarget target) {
         AndroidAppTarget.Keystore keystore = target.keystore
         if (keystore != null) {
-            File storeDir = new File(".okbuck/keystore/${target.identifier.replaceAll(':', '_')}")
+            File storeDir = target.project.rootProject.file(".okbuck/keystore/${target.identifier.replaceAll(':', '_')}")
```
kageiit: you can do target.rootProject directly

buildSrc/src/main/java/com/uber/okbuck/core/util/LintUtil.java
```diff
@@ -73,7 +73,7 @@ private static String getLintwConfigName(Project project, File config) {
 
     @SuppressWarnings("ResultOfMethodCallIgnored")
     public static synchronized String getLintwConfigRule(Project project, File config) {
-        File configFile = new File(LINT_DEPS_CACHE + "/" + getLintwConfigName(project, config));
+        File configFile = project.getRootProject().file(LINT_DEPS_CACHE + "/" + getLintwConfigName(project, config));
```
kageiit: project.rootProject

seanabraham: This is in java not groovy

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/KeystoreRuleComposer.groovy
seanabraham: Whoops

seanabraham:
Fixed up for comment, PTAL @kageiit

buildSrc/src/main/java/com/uber/okbuck/core/util/LintUtil.java
kageiit: ah ok :)

buildSrc/src/main/java/com/uber/okbuck/wrapper/BuckWrapperTask.java
```diff
@@ -52,6 +55,19 @@ void installWrapper() {
         }
     }
 
+    private static String toWatchmanIgnoredDirs(Set<String> ignoredDirs) {
+        if (ignoredDirs.isEmpty()) {
+            return "";
+        }
+
+        String ignore_exprs = ignoredDirs
```
kageiit: camelCase ignore_exprs

buckw
```diff
@@ -55,7 +55,7 @@ DEFAULT_BUCK_INSTALL_DIR="$HOME/.gradle/caches/okbuck/buck"
 CUSTOM_BUCK_REPO=""
 CUSTOM_REMOTE_NAME="custom"
 OKBUCK_SUCCESS="$WORKING_DIR/build/okbuck.success"
-OKBUCK_DIR=".okbuck"
+OKBUCK_DIR="$WORKING_DIR/.okbuck"
```
kageiit: Btw im not sure if this will work as expected. This assumes that the workingdir is always the project root.

kageiit: `.okbuck` should be located next to buckw otherwise the paths to rules will not work

buildSrc/src/main/java/com/uber/okbuck/wrapper/BuckWrapperTask.java
seanabraham: I was copying the pattern from `toWatchmanMatchers` below. Shall I update that as well? Not sure why snake_case was being used in the first place.

buckw
```diff
@@ -75,20 +75,26 @@ getChanges ( ) {
...
+        ["not",
+            ["dirname", ".okbuck"]
         ]
     ],
     "fields": ["name"]
 }]
 EOT`
     SOURCE_ROOTS=`watchman --output-encoding=json --no-pretty -j 2>&1 <<-EOT
-["query", ".", {
+["query", "$WORKING_DIR", {
```
raviagarwal7: I think you would want to have the `$WORKING_DIR` as the root dir as excludes and includes will be added relative to that.

buckw
seanabraham: When okbuck is invoked it is invoked via:
`$WORKING_DIR/gradlew -p $WORKING_DIR okbuck -Dokbuck.wrapper=true $EXTRA_OKBUCK_ARGS`
So currently, it is guaranteed that $WORKING_DIR is your project root (internally we have this rejiggered a little bit)

seanabraham: Why does `.okbuck` need to be next to buckw? Shouldn't it be next to the other folders with generated BUCK within them?

kageiit: nvm i forgot we set the gradle working dir :)

seanabraham:
@kageiit + @raviagarwal7 I added one more change so that when installing a custom buck repo, `buckw` more accurately `cd`s back to the directory you came from. If you haven't modified your working directory so that it's still `pwd`, this shouldn't change anything but this will better protect in the case that someone modifies `buckw` like we have if their working directory is different from `pwd`.

buckw
seanabraham: I'm not sure I totally follow. By excludes/includes do you mean in the context of watchman?

If need be, you can specify a `relative_root` for watchman in the case that you'd want `$WORKING_DIR` to differ from the watched root like I mentioned in the PR summary. Watchman includes/excludes are evaluated relative to the relative_root if specified (https://facebook.github.io/watchman/docs/file-query.html#relative-roots). I'm not making that change right now but it should be possible.

PR #421, opened 2017-04-12
Updated readme with the new android-ndk location

cesarferreira:
Installing as currently stated in the `readme` results in:

```
Error: No available formula with the name "android-ndk"
It was migrated from homebrew/core to caskroom/cask/android-sdk.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ OkBuck is a gradle plugin that lets developers utilize the [Buck](https://buckbu
 brew update
 
 # Required to build and use buck
-brew install android-ndk ant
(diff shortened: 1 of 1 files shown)


kageiit:
I would actually prefer to get rid of the disparity entirely and just put links to the installations. i.e no more Mac OSX/ Linux instructions

jmfayard:
I add the problem and ended up not using brew

The standard procedure to install the NDK via android studio is trivial enough

https://developer.android.com/ndk/guides/index.html

kageiit:
Cool. Closing this for now. README will be updated in a separate change

PR #422, opened 2017-04-14
Fixed transform and added possibility to inject new jars on the fly

malbano:
Fixes:
 - GenRule as executable was never called
 - Updating transform-cli was not working due to wrong transform jar referenced in TransformUtil

Added:
 - Support for new jars in JarsTransformOutputProvider. When invoked without an absolute path, the output folder is automatically appended.


buildSrc/src/main/java/com/uber/okbuck/core/util/TransformUtil.java
```diff
@@ -15,7 +15,7 @@
     public static final String CONFIGURATION_TRANSFORM = "transform";
     private static final String TRANSFORM_FOLDER = "transform/";
     private static final String TRANSFORM_BUCK_FILE = "BUCK_FILE";
-    private static final String TRANSFORM_JAR = "transform-cli-1.0.0.jar";
+    private static final String TRANSFORM_JAR = "transform-cli.jar";
```
kageiit: we actually want the version here to that anyone using this jar will invalidate their existing cache and rules. Please go for v1.1.0 or something

PR #424, opened 2017-04-17
Support invoking okbuck from a different directory

seanabraham:
This introduces the concept of the buck root directory to okbuck. The buck root directory is where buck itself will be invoked from, therefore all the build rules need to be specified from this root. OkBuck still runs relative to the specified gradle working directory. 

I'm not sure if `project.ext` was the best place to store the root directory, as I'm a bit unfamiliar with gradle plugin best practices, open to suggestions (as well as naming suggestions, etc.).

In most cases I plumbed the `Target` through, also not sure if this is most appropriate but since it is available as the source of most of the build rules anyway it felt okay.

@kageiit

### DIFF
--- a/buildSrc/src/main/groovy/com/uber/okbuck/OkBuckGradlePlugin.groovy
+++ b/buildSrc/src/main/groovy/com/uber/okbuck/OkBuckGradlePlugin.groovy
@@ -92,6 +92,9 @@ class OkBuckGradlePlugin implements Plugin<Project> {
         RetrolambdaExtension retrolambda = okbuckExt.extensions.create(RETROLAMBDA, RetrolambdaExtension)
         okbuckExt.extensions.create(TRANSFORM, TransformExtension)
 
+        project.ext.buckRootDirectory = System.getProperty("okbuck.buckRoot") != null ?
+                            new File(System.getProperty("okbuck.buckRoot")) : project.getRootDir();
+
         // Create configurations
         project.configurations.maybeCreate(TransformUtil.CONFIGURATION_TRANSFORM)
         Configuration externalOkbuck = project.configurations.maybeCreate(CONFIGURATION_EXTERNAL)
--- a/buildSrc/src/main/groovy/com/uber/okbuck/composer/groovy/GroovyLibraryRuleComposer.groovy
+++ b/buildSrc/src/main/groovy/com/uber/okbuck/composer/groovy/GroovyLibraryRuleComposer.groovy
@@ -14,16 +14,16 @@ final class GroovyLibraryRuleComposer extends JvmBuckRuleComposer {
 
     static GroovyLibraryRule compose(GroovyLibTarget target) {
         List<String> deps = []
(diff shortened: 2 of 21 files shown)


seanabraham:
This should currently work like normal, but I'm using it with an amended `buckw` which passes the root dir when invoking okbuck via `-Dokbuck.buckRoot=$ROOT_DIR`. I could add that to the standard `buckw` as well and make the directory passed in default to `pwd`, the same as `WORKING_DIR`

buildSrc/src/main/groovy/com/uber/okbuck/OkBuckGradlePlugin.groovy
```diff
@@ -6,23 +6,8 @@ import com.uber.okbuck.core.model.base.ProjectType
...
-import com.uber.okbuck.core.util.TransformUtil
-import com.uber.okbuck.extension.ExperimentalExtension
-import com.uber.okbuck.extension.IntellijExtension
-import com.uber.okbuck.extension.LintExtension
-import com.uber.okbuck.extension.OkBuckExtension
-import com.uber.okbuck.extension.RetrolambdaExtension
-import com.uber.okbuck.extension.TestExtension
-import com.uber.okbuck.extension.TransformExtension
-import com.uber.okbuck.extension.WrapperExtension
+import com.uber.okbuck.core.util.*
```
seanabraham: This was automatically done by the IDE. I can revert if desired.

kageiit: yes please revert all wildcard imports

seanabraham: 👍  Done

seanabraham:
The test failure looks spurious but I'm not able to kick off another in the UI, I can push again if need be though

seanabraham:
Hm, one thing I realized is that this doesn't properly support buckconfig generation yet, that should be pretty straightforward though I believe. Might be a good time to rethink `.buckconfig.local` generation as well

buildSrc/src/main/groovy/com/uber/okbuck/composer/groovy/GroovyLibraryRuleComposer.groovy
```diff
@@ -14,16 +14,16 @@ final class GroovyLibraryRuleComposer extends JvmBuckRuleComposer {
 
     static GroovyLibraryRule compose(GroovyLibTarget target) {
         List<String> deps = []
-        deps.addAll(external(target.main.externalDeps))
+        deps.addAll(external(target.main.externalDeps, target))
```
kageiit: these static methods should not be getting an instance of target passed into them. There is too much leakage of info here. Lets talk over slack on how we can simplify this

kageiit:
Talked with @seanabraham offline and this is not needed anymore

PR #425, opened 2017-04-18
Skip okbuck if we just want to kill the buck daemon.

runningcode:


buckw
```diff
@@ -191,6 +191,12 @@ setupBuckRun ( ) {
     setupBuckBinary
 }
 
+# Do not pass go, do not collect $200 go straight to kill - Skip okbuck if we just want to kill the buck daemon.
+if [[ "kill" == "$1" ]];
```
kageiit: Can we move this to a method declaration? We can handle all the cases in '$1' explicitly then

runningcode: move it in to `setupBuckRun()` or its own method?

kageiit: separate method. Something like

```sh
handleCommands ( ) {
    if [[ "$1" == "kill" ]]; then # Handle kill command
    elif # handle other commands and so on
    fi
}
```

runningcode:
Ok, updated to use a function. My bash script foo isn't very good so let me know if there's a better way to do this.

buckw
```diff
@@ -191,7 +191,16 @@ setupBuckRun ( ) {
...
 }
 
-setupBuckRun
+# Handle parameters and flags
+handleParams ( ) {
+    # Go directly to kill. Do not run okbuck, do not collect $200.
+    if [[ "kill" == $1 ]]; then
+        SKIP_OKBUCK=true
+    fi
+    setupBuckRun
```
kageiit: setupBuckRun should go just below the call to handleParams

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Fri Mar 31 15:13:29 PDT 2017
+#Sat Apr 15 01:58:29 PDT 2017
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.5-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-3.5-all.zip
```
kageiit: Do not need to change this

runningcode: Whoops, I think when the IDE asked if i wanted to use "distribution with sources" I said yes. We don't want the sources?

kageiit: It does not work ever since they generate an API jar for Gradle now. No point bloating up downloads in Travis and ppl working on it. Please change it back

kageiit: It does not work ever since they generate an API jar for Gradle now. No point bloating up downloads in Travis and ppl working on it. Please change it back

buckw
```diff
@@ -191,7 +191,16 @@ setupBuckRun ( ) {
     setupBuckBinary
 }
 
-setupBuckRun
+# Handle parameters and flags
+handleParams ( ) {
+    # Go directly to kill. Do not run okbuck, do not collect $200.
```
cwoodwar6: 👏

buckw
kageiit: Can you move this below `handleParams "$@"`

buckw
kageiit: whats $200?

runningcode: https://en.wikipedia.org/wiki/Do_not_pass_Go._Do_not_collect_$200.

kageiit: That is cool, but please remove it :)

Prefer keeping it clean

gradle/wrapper/gradle-wrapper.properties
kageiit: Can you switch this back to bin as well? The rest looks good

runningcode:
Alright, updated to remove $200 and use the binary distribution of the gradle wrapper.

seanabraham:
@kageiit this is "$200" FYI https://en.wikipedia.org/wiki/Do_not_pass_Go._Do_not_collect_$200

http://vignette3.wikia.nocookie.net/monopoly/images/b/b5/Go_to_Jail.jpg/revision/latest?cb=20140526164909

😛

PR #426, opened 2017-04-18
Add option to easily publish to maven repo

seanabraham:
This just adds a few functions so that it's easy to publish to a maven repo with `./gradlew publish` (some of the logic from https://github.com/chrisbanes/gradle-mvn-push/blob/master/gradle-mvn-push.gradle), probably for internal releases.

`./gradlew bintrayUpload` should still work fine as before

cc @kageiit

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -52,6 +52,19 @@ version = publishVersion
 def siteUrl = 'https://github.com/uber/okbuck'
 def gitUrl = 'https://github.com/uber/okbuck.git'
 
+def getReleaseRepositoryUrl() {
+  return hasProperty('RELEASE_REPOSITORY_URL') ? RELEASE_REPOSITORY_URL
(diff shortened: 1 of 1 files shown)


buildSrc/build.gradle
```diff
@@ -52,6 +52,19 @@ version = publishVersion
 def siteUrl = 'https://github.com/uber/okbuck'
 def gitUrl = 'https://github.com/uber/okbuck.git'
 
+def getReleaseRepositoryUrl() {
+  return hasProperty('RELEASE_REPOSITORY_URL') ? RELEASE_REPOSITORY_URL
+      : "https://oss.sonatype.org/service/local/staging/deploy/maven2/"
```
kageiit: We do not want to deploy snapshots here. We already have bintray which picks up our changes. If you want it purely for internal use, feel free to add this in a separate gradle script and use it via an apply block.

PR #427, opened 2017-04-22
Add support for downloading 22-25 API robolectric runtime dependencies.

raviagarwal7:
Add below to DEFS file to support robolectric 3.3.2

```
original_robolectric_test = robolectric_test
def robolectric_test(
    name,
    force_final_resource_ids=false,
    **kwargs
    ):
    original_robolectric_test(
      name=name,
      force_final_resource_ids=false,
      **kwargs
    )
```
Buck changes will be merged in - https://github.com/facebook/buck/pull/1305


kageiit:
What about versions of robolectric itself? I still see we use the 3.0 artifacts for those?

raviagarwal7:
do you mean shadows-core versions? 3.3.2 ships with only on version of [shadows-core](https://repo1.maven.org/maven2/org/robolectric/shadows-core/3.3.2/), unlike [3.0](https://repo1.maven.org/maven2/org/robolectric/shadows-core/3.0/)

Still keeping the 3.0 around for backwards compatibility.

PR #433, opened 2017-04-26
Do not allow okbuck to be run without okbuck.wrapper set to true.

runningcode:


buildSrc/src/main/groovy/com/uber/okbuck/OkBuckGradlePlugin.groovy
```diff
@@ -125,6 +125,12 @@ class OkBuckGradlePlugin implements Plugin<Project> {
                 [cacheName, project.configurations.maybeCreate("${cacheName}ExtraDepCache")]
             }
 
+            setupOkbuck.doFirst {
+                if (!System.getProperty("okbuck.wrapper", "false").toBoolean()) {
+                    throw new IllegalArgumentException("Okbuck task cannot be invoked without okbuck.wrapper set to true")
```
kageiit: `throw new IllegalArgumentException("Okbuck cannot be invoked without okbuck.wrapper set to true. Prefer to use buckw instead")`

runningcode:
Updated!

PR #436, opened 2017-04-28
Move okbuck task to its own gradle task class.

runningcode:


buildSrc/src/main/java/com/uber/okbuck/core/task/OkBuckTask.java
```diff
@@ -0,0 +1,107 @@
...
+import java.util.stream.Collectors;
+
+import static com.uber.okbuck.OkBuckGradlePlugin.OKBUCK_DEFS;
+
+public class OkBuckTask extends DefaultTask {
+
+  @Input public OkBuckExtension okBuckExtension;
+
+  @TaskAction
+  void doOkBuckTask() {
```
kageiit: can you rename this method to `okbuck`

runningcode: done!

buildSrc/src/main/java/com/uber/okbuck/core/task/OkBuckTask.java
```diff
@@ -0,0 +1,107 @@
...
+            kotlinDeps != null ? kotlinDeps.getLeft(): null,
+            kotlinDeps != null ? kotlinDeps.getRight(): null);
+  }
+
+  @Override public String getGroup() {
+    return OkBuckGradlePlugin.GROUP;
+  }
+
+  @Override public String getDescription() {
+    return "Setup okbuck cache and dependencies";
```
kageiit: This should be handled by the `setupOkbuck` Task. root okbuck task description should be more like "okbuck task for the root project"

buildSrc/src/main/groovy/com/uber/okbuck/OkBuckGradlePlugin.groovy
```diff
@@ -99,11 +93,11 @@ class OkBuckGradlePlugin implements Plugin<Project> {
         // Create tasks
         Task setupOkbuck = project.task('setupOkbuck')
         setupOkbuck.setGroup(GROUP)
-        setupOkbuck.setDescription("Setup okbuck cache and dependencies")
+        setupOkbuck.setDescription()
 
-        Task okBuck = project.task(OKBUCK)
-        okBuck.setGroup(GROUP)
-        okBuck.setDescription("Generate BUCK files")
+        Task okBuck = project.tasks.create(OKBUCK, OkBuckTask, {
```
kageiit: okbuck task is not safe to be an incremental task. You should set it as never up to date (which `doLast` does as a side effect). When it is truly incremental, we can lift that restriction

buildSrc/src/main/groovy/com/uber/okbuck/OkBuckGradlePlugin.groovy
```diff
@@ -99,11 +93,11 @@ class OkBuckGradlePlugin implements Plugin<Project> {
         // Create tasks
         Task setupOkbuck = project.task('setupOkbuck')
         setupOkbuck.setGroup(GROUP)
-        setupOkbuck.setDescription("Setup okbuck cache and dependencies")
```
kageiit: Please undo this

buildSrc/src/main/java/com/uber/okbuck/core/task/OkBuckTask.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  @InputFile
+  public File okbuckDefs() {
+    return getProject().file(OKBUCK_DEFS);
```
kageiit: This should be an output file?

buildSrc/src/main/java/com/uber/okbuck/core/task/OkBuckTask.java
```diff
@@ -0,0 +1,107 @@
...
+  }
+
+  @OutputFile
+  public File dotBuckConfig() {
+    return getProject().file(".buckconfig");
```
kageiit: This is only generated if it does not exist so I am not sure we should call this an outputfile. We can check if we can remove the generation of this file entirely as the contents are empty anyways

buildSrc/src/main/groovy/com/uber/okbuck/OkBuckGradlePlugin.groovy
runningcode: whoops, i copied the wrong description

buildSrc/src/main/java/com/uber/okbuck/core/task/OkBuckTask.java
runningcode: updated!

buildSrc/src/main/java/com/uber/okbuck/core/task/OkBuckTask.java
runningcode: yup

buildSrc/src/main/groovy/com/uber/okbuck/OkBuckGradlePlugin.groovy
runningcode: ok, i've added this to the constructor to ensure that the up to date check always returns false
`getOutputs().upToDateWhen(Specs.satisfyNone());`

buildSrc/src/main/java/com/uber/okbuck/core/task/OkBuckTask.java
runningcode: ah, i've removed the `@OutputFile` annotation. do you think it should be an `@InputFile`?
ive disabled the up to date check so i think it shouldn't matter.

PR #437, opened 2017-05-04
Use stream instead of parallelStream

runningcode:
The overhead of creating threads for very small tasks might not be
faster than performing the same tasks serially.

I ran a benchmark using this commit against our repo. I killed the
gradle daemon and then ran the okbuck task 3 times.

Before: using parallel()
20.353 secs
6.644 secs
5.89 secs

After: using stream()
18.539 secs
6.316 secs
5.613 secs

The results were quite similar with and without parallel stream. I'm
wondering if there is any noticeable difference with even larger projects.

### DIFF
--- a/buildSrc/src/main/groovy/com/uber/okbuck/composer/base/BuckRuleComposer.java
+++ b/buildSrc/src/main/groovy/com/uber/okbuck/composer/base/BuckRuleComposer.java
@@ -14,13 +14,13 @@ public abstract class BuckRuleComposer {
     private static final String SEPARATOR = ":";
 
     public static Set<String> external(final Set<String> deps) {
-        return deps.parallelStream()
+        return deps.stream()
                 .map(BuckRuleComposer::fileRule)
                 .collect(Collectors.toSet());
     }
 
     public static Set<String> externalApt(final Set<String> deps) {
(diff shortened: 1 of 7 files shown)


kageiit:
Ran it on our codebase

Before:
```
Total time: 53.786 secs
Total time: 34.565 secs
Total time: 32.101 secs
```

After
```
Total time: 51.386 secs
Total time: 34.451 secs
Total time: 30.859 secs
```

Looks good

buildSrc/build.gradle
```diff
@@ -46,7 +46,7 @@ task sourcesJar(type: Jar) {
     classifier = 'sources'
 }
 
-def publishVersion = '0.20.7'
+def publishVersion = '0.20.8-SNAPSHOT'
```
kageiit: Can you revert this?

runningcode: Isn't it typical to bump the version to the next release snapshot? This makes it easier to test changes locally.

kageiit: you can do so if you want to locally. Even if we wanted to start doing that, it should be a separate change

runningcode:
Removed snapshot version!

PR #438, opened 2017-05-04
Add empty resource rule if using resource union.

runningcode:
This ensures that an R file is created for that module even if there are
no resources. Gradle always generates an R file for each module even if
there are no resources.

### DIFF
--- a/libraries/emptyResourceLibrary/build.gradle
+++ b/libraries/emptyResourceLibrary/build.gradle
@@ -1,0 +1,6 @@
+apply plugin: 'com.android.library'
+
+dependencies {
+    compile project(':libraries:emptylibrary')
+    compile deps.support.appCompat
+}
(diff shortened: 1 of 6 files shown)


libraries/emptyResourceLibrary/build.gradle
```diff
@@ -0,0 +1,10 @@
+apply plugin: 'com.android.library'
+apply plugin: 'me.tatarka.retrolambda'
+
+android {
```
kageiit: you dont need the empty android block

libraries/emptyResourceLibrary/build.gradle
```diff
@@ -0,0 +1,10 @@
+apply plugin: 'com.android.library'
+apply plugin: 'me.tatarka.retrolambda'
```
kageiit: remove retrolambda?

runningcode:
Removed retrolambda and removed empty android block.

PR #439, opened 2017-05-05
Remove unused parameters and clean up some code.

runningcode:


buildSrc/src/main/groovy/com/uber/okbuck/generator/BuckFileGenerator.groovy
```diff
@@ -97,7 +94,7 @@ final class BuckFileGenerator {
                     }
                     break
                 default:
-                    break
+                    throw new IllegalArgumentException("ProjectType not handled: $projectType")
```
kageiit: This will break users who forget to set buck projects in the extension. That's a breaking API change. What's the reason for this?

runningcode: My goal was to make it easier when adding a new `ProjectType` to figure out all the places that it needs to be added to in the code.

I didn't realize this would crash in other ways, are you saying that if incompatible projects aren't explicitly excluded from `buckProjects` in the `OkbuckExtension` that this method will crash here?

From my experience, incompatible projects will just crash further down the pipeline if they don't already crash here. Kotlin Android projects are one example of this so I don't think it will change much.

I'm happy to add the following code if it would help this problem:
```
UNKOWN:
  break;
default:
  throw new IllegalArgumentException("...");
```

kageiit: Please change it to print a warning instead of throwing an exception

kageiit: Any plans to make this change? I can close it otherwise

runningcode: I strongly think that throwing an exception is the best option here. Otherwise when a developer adds a new ProjectType to the enum, it will silently fail. This took me a while to debug when I added a new `KotlinAndroidModule` as a project type.

kageiit: An exception is good only if we can offer some mitigation. When a project type is not handled, we should put a message in the exception to excldue the project using `okbuck.buckProjects` extension. Otherwise it is not actionable.

I am ok taking the exception, if we make the error message clearer

kageiit: Will address this in a followup myself so I can merge this

PR #441, opened 2017-05-05
Move generate dotbuckconfig local in to its own task.

runningcode:


kageiit:
we actually had it that way before and moved it all into setupOkbuck because tasks in the same project cannot be parallelized

We should also never call `.execute()` directly when creating a task. That is basically like calling a regular method at that point and no need to make something a task.

Feel free to move the logic into a separate method/utility class but this should not be a separate task

PR #445, opened 2017-05-18
Preserve tools:overrideLibrary in <uses-sdk>

marcinkwiatkowski:
okbuck currently replaces whole <uses-sdk> tag and overrides potential tools:overrideLibrary

### DIFF
--- a/buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidTarget.groovy
+++ b/buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidTarget.groovy
@@ -240,14 +240,16 @@ abstract class AndroidTarget extends JavaLibTarget {
     }
 
     String processManifestXml(GPathResult manifestXml) {
-        def sdkNode = {
-            'uses-sdk'('android:minSdkVersion': String.valueOf(minSdk),
-                    'android:targetSdkVersion': String.valueOf(targetSdk)) {}
-        }
         if (manifestXml.'uses-sdk'.size() == 0) {
-            manifestXml.appendNode(sdkNode)
+            manifestXml.appendNode {
+                'uses-sdk'('android:minSdkVersion': String.valueOf(minSdk), 'android:targetSdkVersion': String.valueOf(targetSdk)) {}
+            }
         } else {
-            manifestXml.'uses-sdk'.replaceNode(sdkNode)
+            def overrideLibrary = manifestXml.'uses-sdk'.@'tools:overrideLibrary'
+            manifestXml.'uses-sdk'.replaceNode {
+                'uses-sdk'('android:minSdkVersion': String.valueOf(minSdk), 'android:targetSdkVersion': String.valueOf(targetSdk),
+                    'xmlns:tools':'http://schemas.android.com/tools', 'tools:overrideLibrary': overrideLibrary) {}
+            }
         }
 
         def builder = new StreamingMarkupBuilder()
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidTarget.groovy
```diff
@@ -240,14 +240,16 @@ abstract class AndroidTarget extends JavaLibTarget {
...
-                    'android:targetSdkVersion': String.valueOf(targetSdk)) {}
-        }
         if (manifestXml.'uses-sdk'.size() == 0) {
-            manifestXml.appendNode(sdkNode)
+            manifestXml.appendNode {
+                'uses-sdk'('android:minSdkVersion': String.valueOf(minSdk), 'android:targetSdkVersion': String.valueOf(targetSdk)) {}
+            }
         } else {
-            manifestXml.'uses-sdk'.replaceNode(sdkNode)
+            def overrideLibrary = manifestXml.'uses-sdk'.@'tools:overrideLibrary'
```
kageiit: Can you add this only if it is declared/exists?

This should be explicitly declared in the base manifest before we can add it

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidTarget.groovy
```diff
@@ -240,14 +240,20 @@ abstract class AndroidTarget extends JavaLibTarget {
     }
 
     String processManifestXml(GPathResult manifestXml) {
-        def sdkNode = {
-            'uses-sdk'('android:minSdkVersion': String.valueOf(minSdk),
-                    'android:targetSdkVersion': String.valueOf(targetSdk)) {}
-        }
         if (manifestXml.'uses-sdk'.size() == 0) {
-            manifestXml.appendNode(sdkNode)
+            manifestXml.appendNode {
```
kageiit: lets bring back `sdkNode` variable, add the override inside that block (if needed) and keep the old logic for adding the node as it was

kageiit:
Can you please also sign the CLA?

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidTarget.groovy
marcinkwiatkowski: hmm, not sure if I can have logic there

kageiit: Thats nice. We can make the logic simpler here by constructing the sdkNode correctly all the time i.e read any and all existing attrs if the node exists, and add our modifications on top when creating the new node to always replace existing.

How does that sound? Then it is no loner specific to `tools:overrideLibrary` and will work for all attributes

marcinkwiatkowski: you mean something like:
```
        def sdkAttributes = manifestXml.'uses-sdk'.'**'*.attributes()[0]
        sdkAttributes['{http://schemas.android.com/apk/res/android}minSdkVersion'] = String.valueOf(minSdk)
        sdkAttributes['{http://schemas.android.com/apk/res/android}targetSdkVersion'] = String.valueOf(targetSdk)
        def sdkNode = { 'uses-sdk'(sdkAttributes) }
```

kageiit: Yep :)

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidTarget.groovy
```diff
@@ -260,6 +257,16 @@ abstract class AndroidTarget extends JavaLibTarget {
                 .replaceFirst('<manifest ', '<manifest xmlns:android="http://schemas.android.com/apk/res/android" ')
     }
 
+    private groovy.lang.Closure getSdkNode(GPathResult manifestXml) {
```
kageiit: no need for a fullt qualified import

This can also be static

marcinkwiatkowski: hmm, maybe that should go to some utils class?

PR #447, opened 2017-05-24
Do not allow snapshot/dynamic dependencies in okbuck

kageiit:
This makes it safer for the sake of hermetic builds by surfacing any dynamically versioned dependencies to users

- By default a warning is printed, but it can be made an error by setting `failOnChangingDependencies` in the `okbuck` extension. This behavior will be made to always error on a subsequent release
- Also updated gradle to latest


marcinkwiatkowski:
👍

PR #448, opened 2017-05-25
Fix tools namespace in merged manifest

marcinkwiatkowski:


buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidTarget.groovy
```diff
@@ -252,9 +252,11 @@ abstract class AndroidTarget extends JavaLibTarget {
         def builder = new StreamingMarkupBuilder()
         builder.setUseDoubleQuotes(true)
         return (builder.bind {
+            mkp.declareNamespace(tools:'http://schemas.android.com/tools')
```
kageiit: This should be declared in the input manifest

kageiit:
Seems like CI failed

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidTarget.groovy
```diff
@@ -254,9 +251,8 @@ abstract class AndroidTarget extends JavaLibTarget {
         return (builder.bind {
             mkp.yield manifestXml
         } as String)
-                .replaceAll('\\{http://schemas.android.com/apk/res/android\\}', 'android:')
-                .replaceAll('xmlns:android="http://schemas.android.com/apk/res/android"', '')
-                .replaceFirst('<manifest ', '<manifest xmlns:android="http://schemas.android.com/apk/res/android" ')
+            .replaceAll('xmlns:android="http://schemas.android.com/apk/res/android"', '')
```
kageiit: why do we still need these?

marcinkwiatkowski: This is to handle manifests that don't declare android namespace. The build was previously falling when merging https://github.com/uber/okbuck/blob/master/libraries/emptyResourceLibrary/src/main/AndroidManifest.xml

PR #449, opened 2017-05-26
Switch the ordering of build config fields.

cwoodwar6:
In the below example, we would expect the final value to be false, but instead it is true.

```
android {
  defaultConfig {
    buildConfigField "boolean", "VALUE", "true"
  }

  buildTypes {
    debug {
      buildConfigField "boolean", "VALUE", "false"
    }
  }
}
```
With the order switched, it uses the expected value of false instead of true. In Gradle, the result is always false. Gradle prefers to use the build type value over the default.

### DIFF
--- a/buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidTarget.groovy
+++ b/buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidTarget.groovy
@@ -177,11 +177,11 @@ abstract class AndroidTarget extends JavaLibTarget {
 
         Map<String, ClassField> extraBuildConfig = [:]
 
-        baseVariant.buildType.buildConfigFields.collect { String key, ClassField classField ->
+        baseVariant.mergedFlavor.buildConfigFields.collect { String key, ClassField classField ->
             extraBuildConfig.put(key, classField)
         }
 
-        baseVariant.mergedFlavor.buildConfigFields.collect { String key, ClassField classField ->
+        baseVariant.buildType.buildConfigFields.collect { String key, ClassField classField ->
             extraBuildConfig.put(key, classField)
         }
 


buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidTarget.groovy
```diff
@@ -177,11 +177,11 @@ abstract class AndroidTarget extends JavaLibTarget {
 
         Map<String, ClassField> extraBuildConfig = [:]
 
-        baseVariant.buildType.buildConfigFields.collect { String key, ClassField classField ->
+        baseVariant.mergedFlavor.buildConfigFields.collect { String key, ClassField classField ->
```
kageiit: what if you added a buildconfig field for a product flavor?

Can you check what the output is in that case?

cwoodwar6:
```
android {
  defaultConfig {
    buildConfigField "Integer", "VALUE", "1"
  }

  buildTypes {
    debug {
      buildConfigField "Integer", "VALUE", "2"
    }
  }

  productFlavors {
    full {
      buildConfigField "Integer", "VALUE", "3"
    }
  }
}
```

In the above case, my change will cause the end result to be 2 when it should be 3. Good catch. However, without the flavor, the behavior isn't as expected without this change.

kageiit: Yes. I think we should update it fallback correctly in all the cases

cwoodwar6: My assumption seems incorrect. Gradle seems to give precedence to buildTypes. So 2 should be the correct value in both cases.

PR #454, opened 2017-06-19
Bump to gradle 4.0 stable.

runningcode:


build.gradle
```diff
@@ -40,6 +40,10 @@ subprojects { project ->
     }
 }
 
+task wrapper(type: Wrapper) {
```
kageiit: Lets leave this out. Its easier to just invoke the version on the commandline usually

kageiit: Will clean this up after

PR #464, opened 2017-06-27
Setup Kotlin home whenever the kotlin gradle plugin is on the classpath.

runningcode:
Previously we should only check if a module was a kotlin module which
didn't account for kotlin-android modules.

### DIFF
--- a/buildSrc/src/main/java/com/uber/okbuck/core/task/OkBuckTask.java
+++ b/buildSrc/src/main/java/com/uber/okbuck/core/task/OkBuckTask.java
@@ -13,7 +13,6 @@ import com.uber.okbuck.core.util.ProjectUtil;
 import com.uber.okbuck.extension.OkBuckExtension;
 import com.uber.okbuck.generator.DotBuckConfigLocalGenerator;
 
-import org.apache.commons.lang3.tuple.Pair;
 import org.gradle.api.DefaultTask;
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/java/com/uber/okbuck/core/task/OkBuckTask.java
```diff
@@ -46,7 +45,7 @@ void okbuck() {
     }
 
     // Fetch Kotlin support deps if needed
-    boolean hasKotlinLib = okBuckExtension.buckProjects.parallelStream().anyMatch(project -> ProjectUtil.getType(project) == ProjectType.KOTLIN_LIB);
+    boolean hasKotlinLib = ProjectUtil.hasArtifactInClasspath(getProject(), KotlinUtil.KOTLIN_GROUP, KotlinUtil.KOTLIN_GRADLE_MODULE);
```
kageiit: Can we move this utility method into KotlinUtil instead? Its weird to put this on project util if its not reused anywhere else

kageiit: We can just call it KotlinUtil.hasKotlin(project) then as the constants can remain private

runningcode:
Code looks much cleaner! :)

PR #470, opened 2017-07-11
Skip OkBuck if we just want to run buck options.

cwoodwar6:
We shouldn't need to run the okbuck task if all we want to do is run `--help` on a buck command.

### DIFF
--- a/buckw
+++ b/buckw
@@ -4,7 +4,7 @@
 ##
 ##  Buck wrapper script to invoke okbuck when needed, before running buck
 ##
-##  Created by OkBuck Gradle Plugin on : Thu Jul 13 10:40:45 PDT 2017
+##  Created by OkBuck Gradle Plugin on : Fri Jul 14 10:17:41 PDT 2017
 ##
 #########################################################################
 
@@ -213,10 +213,10 @@ setupBuckRun ( ) {
 
 # Handle parameters and flags
 handleParams ( ) {
-    # Go directly to kill. Do not run okbuck.
-    if [[ "kill" == $1 ]]; then
-        SKIP_OKBUCK=true
-    fi
+   # Go directly to the kill command, help command, or --help option. Do not run okbuck.
+   if [[ "kill" == $1 || "help" == $1 || $@ == *"--help"* || $@ == *"-h"* ]]; then
+      SKIP_OKBUCK=true
+   fi
 }
 
 handleParams "$@"
--- a/buildSrc/src/main/resources/com/uber/okbuck/core/util/wrapper/BUCKW_TEMPLATE
+++ b/buildSrc/src/main/resources/com/uber/okbuck/core/util/wrapper/BUCKW_TEMPLATE
@@ -204,10 +204,10 @@ setupBuckRun ( ) {
 
 # Handle parameters and flags
 handleParams ( ) {
-    # Go directly to kill. Do not run okbuck.
-    if [[ "kill" == $1 ]]; then
-        SKIP_OKBUCK=true
-    fi
+   # Go directly to the kill command, help command, or --help option. Do not run okbuck.
+   if [[ "kill" == $1 || "help" == $1 || $@ == *"--help"* || $@ == *"-h"* ]]; then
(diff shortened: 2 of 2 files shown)


buckw
```diff
@@ -217,6 +217,9 @@ handleParams ( ) {
     if [[ "kill" == $1 ]]; then
```
kageiit: Lets just make this a composite statement

cwoodwar6: Done.

buckw
```diff
@@ -213,8 +213,8 @@ setupBuckRun ( ) {
 
 # Handle parameters and flags
 handleParams ( ) {
-    # Go directly to kill. Do not run okbuck.
-    if [[ "kill" == $1 ]]; then
+    # Go directly to kill or --help. Do not run okbuck.
+    if [[ "kill" == $1 || "--help" == $2 ]]; then
```
kageiit: I think `--help` need not be the second argument always. Maybe it is best to check if the `$@` matches `--help` instead of doing it on a positional argument?

cwoodwar6: Good point, `buck --help` is an option. I've only used `buck <command> --help`. I also realize you can use just `-h`. There is also `--version` or `-V`. I'll incorporate those.

kageiit: 👍

buckw
```diff
@@ -213,10 +213,15 @@ setupBuckRun ( ) {
 
 # Handle parameters and flags
 handleParams ( ) {
-    # Go directly to kill. Do not run okbuck.
-    if [[ "kill" == $1 ]]; then
-        SKIP_OKBUCK=true
-    fi
+    for arg in "$@"
```
kageiit: It is easier to do

if ($1 == kill) or ( $@ == *"--help"*)

To avoid having to iterate over all the arguments

cwoodwar6: I mostly did it this way for any future argument checking. I'll go ahead and do it your simplified way.

kageiit: We can also use `comm` to see if there is an intersection of arguments we want to check for. See https://stackoverflow.com/a/2696111 for example. That would make it very easy to add new arguments

cwoodwar6: The `comm` command seems to take only files as input. The example uses process substitution to read a process's output as a file. I'm not seeing a way to do it with `comm`?

kageiit: You can use `echo`

<(echo "${Array1[*]}") <(echo "${Array2[*]}")))

cwoodwar6: Ahh fancy. Thanks for the bash hacks.

cwoodwar6: I ended up using `printf` so I could split the arrays with new lines. Otherwise `echo` printed them out as one string instead of a list.

buckw
```diff
@@ -213,10 +213,11 @@ setupBuckRun ( ) {
 
 # Handle parameters and flags
 handleParams ( ) {
-    # Go directly to kill. Do not run okbuck.
-    if [[ "kill" == $1 ]]; then
-        SKIP_OKBUCK=true
-    fi
+    # Go directly to the kill command, help command, or buck options. Do not run okbuck.
+   skip=("kill" "help" "--help" "-h" "--version" "-V")
```
kageiit: I would not put `kill` and `help` here.

because those are valid target names. we should only skip if kill is `$1`

cwoodwar6: Buck itself would catch the error if a command is passed in a position other than `$1`. I couldn't think of a use case where it would be risky to skip if `kill` or `help` is found in another argument.  Especially since `./buckw -V help` is valid and will run the help command but ignore the `-V`.

kageiit:
Wait I know why travis is failing. we use `./buckw --version` as a cheap way to invoke and run okbuck after which we always set `SKIP_OKBUCK=true`. We do this infact on our CI servers as well.

I am not sure if there is any other way to force an okbuck run because @runningcode added a PR to thrown an exception if trying to okbuck from gradlew directly. I think we should keep this functionality as is and you can modify your wrapper locally if you do not need it

cwoodwar6:
Took out `--version` and went back to the simple composite statement.

PR #479, opened 2017-07-20
Fix DependencyCache.fetchSourcesFor() performance regression and incorrect sources jar search logic.

### LINKED ISSUE #473: :setupOkbuck execution time regression (sometimes it takes 1m+)
We're facing non-deterministic regression in `./buckw project` execution time after updating OkBuck from 0.20.4 to 0.22.4.

Expected behavior:

`setupOkbuck` should take 10-30 seconds.

Actual behavior:

`setupOkbuck` takes 50-120 seconds sometimes.

---

Environment: 

- OS: macOS Sierra 10.12.5
- JDK/JVM: Oracle JDK 1.8.0_131
- Gradle: 4.0.1 (happens with 3.5 as well)
- Android Gradle Plugin: 2.3.3
- OkBuck: 0.22.4 (we think regression started somewhere around 0.20.7)

---

Details of our investigation so far:

- We don't see this problem with OkBuck 0.20.4 and Gradle 4.0.1
- First time we've noticed it was May 25 when we've tried to upgrade OkBuck to 0.20.9
- Issue is definitely non-deterministic, but once you catch it, it'll be much more reproducible until you remove Gradle and Buck build folders
- It might be related to changes from #437 ("Use stream instead of parallelStream") which may create bottleneck in some cases.

Example of build log output:

```console
:setupOkbuck (Thread[Task worker,5,main]) started.
:setupOkbuck
Putting task artifact state for task ':setupOkbuck' into context took 0.0 secs.
Executing task ':setupOkbuck' (up-to-date check took 0.0 secs) due to:
  Task has not declared any outputs.
:setupOkbuck (Thread[Task worker,5,main]) completed. Took 1 mins 45.773 secs.
```

@bencodes tried to investigate this back in May 25 and pointed that this was in his build log during long `setupOkbuck` task execution:

```console
> Task :setupOkbuck
Putting task artifact state for task ':setupOkbuck' into context took 0.0 secs.
Executing task ':setupOkbuck' (up-to-date check took 0.0 secs) due to:
  Task has not declared any outputs.
Creating new cache for metadata-2.23/module-artifacts, path /Users/user/.gradle/caches/modules-2/metadata-2.23/module-artifacts.bin, access org.gradle.cache.internal.DefaultCacheAccess@58c0d216
// Nothing in output for long time.
```

We'll try to add more info, but since it's non-deterministic it's really hard to debug and profile.

artem-zinnatullin:
Fixes #473 caused by performance and logic regression in `DependencyCache.fetchSourcesFor()` introduced in PR #423.

This fix reduces `:setupOkbuck` from 1m+ to ~6 seconds in our project 😸 🚀  🎉  

---

#### Performance regression:

When some dependency is stored as `.jar` in near to top-level of the project directory, i.e `project/libs` (which is bad but sometimes that happens), then OkBuck will do `fileTree` lookup **in the project directory** for each such dependency. 

In our case we had 4 jars stored in the project directory and fileTree lookup for sources jar took about 15-25 seconds each (lots of files in `.git`, `.okbuck`, `build`, etc folders) which **resulted in 60 seconds+ execution time** of `:setupOkbuck` task.

#### Logic regression:

OkBuck could find sources jar from wrong library by full `fileTree` lookup in the project directory since most of the jars are named only as `artifactId` and do not contain `groupId` in them. See http://blog.joda.org/2017/04/java-se-9-jpms-module-naming.html (Ownership and Uniqueness).

i.e.:

```console
module-1/libs/spark.core-1.0.jar         # Spark Framework
module-2/libs/spark.core-1.0.jar         # Apache Spark
module-2/libs/spark.core-1.0-sources.jar # Apache Spark
```

OkBuck will use `Apache Spark` sources for `Spark Framework` jar.

---

### Detailed explanation of the fix in this PR

If jar is in the project directory — then try to find sources jar only in the same directory as the jar itself, which avoids both expensive `fileTree` lookup and logic regression.

Otherwise use existing strategy to find sources jar in `jar/../../` directory (works for Gradle/Maven cache).


buildSrc/src/main/groovy/com/uber/okbuck/core/dependency/DependencyCache.groovy
```diff
@@ -109,13 +110,21 @@ class DependencyCache {
...
             if (!sourcesJar.exists()) {
-                def sourceJars = rootProject.fileTree(
-                        dir: dependency.depFile.parentFile.parentFile.absolutePath,
-                        includes: ["**/${sourcesJarName}"]) as List
-                if (sourceJars.size() == 1) {
-                    sourcesJar = sourceJars[0]
-                } else if (sourceJars.size() > 1) {
-                    throw new IllegalStateException("Found multiple source jars: ${sourceJars} for ${dependency}")
+                if (FileUtils.directoryContains(rootProject.projectDir, dependency.depFile)) {
+                    // Jas is in the project directory, try to use sources jar right next to the jar itself.
```
kageiit: Typoe - > Jar

buildSrc/src/main/groovy/com/uber/okbuck/core/dependency/DependencyCache.groovy
```diff
@@ -109,13 +110,21 @@ class DependencyCache {
...
 
             if (!sourcesJar.exists()) {
-                def sourceJars = rootProject.fileTree(
-                        dir: dependency.depFile.parentFile.parentFile.absolutePath,
-                        includes: ["**/${sourcesJarName}"]) as List
-                if (sourceJars.size() == 1) {
-                    sourcesJar = sourceJars[0]
-                } else if (sourceJars.size() > 1) {
-                    throw new IllegalStateException("Found multiple source jars: ${sourceJars} for ${dependency}")
+                if (FileUtils.directoryContains(rootProject.projectDir, dependency.depFile)) {
```
kageiit: Can you check if dependency is local first? There is a property `depenency.isLocal` and this logic should only apply to such dependencies

artem-zinnatullin: Oh, that's actually better!

Looks like it's okay to actually replace `FileUtils.directoryContains()` with `dependency.isLocal`, please correct me if I'm wrong :)

buildSrc/src/main/groovy/com/uber/okbuck/core/dependency/DependencyCache.groovy
artem-zinnatullin: Typoe 😹 

hahaha-classic-3602179

kageiit:
As an integration test, can you add sources to the existing local dependency in the okbuck repo? so we can see it is being picked up correctly

artem-zinnatullin:
The only local dependency I've found in OkBuck is `app/libs/buck-android-support.jar`. 

Do you want me to put fake/real sources jar next to it or put some other example jar with sources next in that directory?

artem-zinnatullin:
I'm also not sure what do you mean by integration test.

Check if sources are visible in IntelliJ OR create automated test that'll run `./buckw project` and check that `.okbuck/caches` contains both `.jar` and `-sources.jar`?

kageiit:
We have a local aar in okbuck
 Travis build everything already.

I'm suggesting you add a sources.jar file right next to the local aar so when Travis runs the tests, it gets picked up

artem-zinnatullin:
Alright, added sources jar for rxscreenshotdetector and also added version of it to the filename to ease release tracking.

`./buckw project` finds and copies `-sources.jar` to the `.okbuck/cache`!

artem-zinnatullin:
Can you please ship v0.23.1 with this fix?

kageiit:
v0.23.1 is live

artem-zinnatullin:
Thanks, appreciate that! ❤️

PR #481, opened 2017-07-27
Generate kotlinc plugin options for Kotlin Android Extensions

cwoodwar6:
Gives OkBuck the ability to generate the kotlinc arguments required for running kotlin android extensions.


buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidLibraryRuleComposer.groovy
```diff
@@ -4,11 +4,14 @@ import com.uber.okbuck.core.model.android.AndroidLibTarget
 import com.uber.okbuck.core.model.android.AndroidTarget
 import com.uber.okbuck.core.model.base.RuleType
 import com.uber.okbuck.core.model.base.Target
+import com.uber.okbuck.core.util.KotlinUtil
 import com.uber.okbuck.core.util.RetrolambdaUtil
 import com.uber.okbuck.rule.android.AndroidLibraryRule
 
 final class AndroidLibraryRuleComposer extends AndroidBuckRuleComposer {
 
+     private static final String PLUGIN_OPTION = "plugin:org.jetbrains.kotlin.android:"
```
artem-zinnatullin: Name looks too generic for `AndroidLibraryRuleComposer` class

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidTarget.groovy
```diff
@@ -94,6 +95,7 @@ abstract class AndroidTarget extends JavaLibTarget {
 
         // Check if kotlin
         isKotlin = project.plugins.hasPlugin(KotlinAndroidPluginWrapper.class)
+        hasKotlinAndroidExtensions = project.plugins.hasPlugin('kotlin-android-extensions')
```
artem-zinnatullin: `KotlinUtil.KOTLIN_ANDROID_EXTENSIONS_MODULE`?

buildSrc/src/main/groovy/com/uber/okbuck/rule/android/AndroidTestRule.groovy
```diff
@@ -40,6 +40,7 @@ final class AndroidTestRule extends AndroidRule {
                 aptDeps,
                 providedDeps,
                 aidlRuleNames,
+                null,
```
kageiit: why can there be no extensions for tests? is it not considered a valid variant?

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidLibraryRuleComposer.groovy
```diff
@@ -46,6 +49,22 @@ final class AndroidLibraryRuleComposer extends AndroidBuckRuleComposer {
...
         }
 
+        List<String> extraKotlincArgs = []
+        if (target.hasKotlinAndroidExtensions) {
+            String options = ""
+            target.getResDirs().each {  String dir ->
+                // :root:module -> /root/module/
+                String module = target.project.path.replace(":", "/") + "/"
+                // src/main/res -> main
+                String variant = dir.substring(dir.indexOf("/") + 1, dir.lastIndexOf("/"))
```
kageiit: why not target.name? that is in fact the correct variant name

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidLibraryRuleComposer.groovy
```diff
@@ -46,6 +49,22 @@ final class AndroidLibraryRuleComposer extends AndroidBuckRuleComposer {
             testTargets.add(":${test(target)}")
         }
 
+        List<String> extraKotlincArgs = []
+        if (target.hasKotlinAndroidExtensions) {
```
kageiit: can we use the same approach for kapt?

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidLibraryRuleComposer.groovy
cwoodwar6: Yeah definitely. Naming is hard, figured I'd come up with a better name before I'm ready to merge.

buildSrc/src/main/groovy/com/uber/okbuck/rule/android/AndroidTestRule.groovy
cwoodwar6: I was actually waiting for your Buck PR to merge so I can verify it was working in tests.

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidLibraryRuleComposer.groovy
cwoodwar6: I'll double check, I think it was defaulting to `debug` in both cases. I need it to be both `main`and `debug`.

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidLibraryRuleComposer.groovy
cwoodwar6: We definitely can, kapt is a plugin to the compiler just how kotlin android extensions is. We'll of course need the changes to buck that run each step involved when doing kapt.

buildSrc/src/main/groovy/com/uber/okbuck/rule/android/AndroidTestRule.groovy
kageiit: Can you not cherry pick and test it locally? Your PR has not merged so far either :)

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidLibraryRuleComposer.groovy
kageiit: I would rather this be coming from the target than the composer. This block should not access anything about the target. The composers should be target api agnostic which is why we pass in all arguments via constructor. This keeps it decoupled and be able to extend to other cases.

Can you create a method in the android target instead that will return a list of kotlinc arguments and we can add logic for various extensions/plugins there instead?

buildSrc/src/main/groovy/com/uber/okbuck/rule/android/AndroidTestRule.groovy
cwoodwar6: I certainly can, but I simply have not yet. We aren't using the extensions in any of our kotlin tests so I haven't had a chance to write a test and use it.

kageiit: Is it hard to add a simple test case? You could just try it in okbuck too. That would serve as an integration test as well so we do not break this flow. I think that should be required as part of the PR to add a simple use case for this flow.

cwoodwar6: I marked the PR as a work in progress for that reason. Wanted to get this out early for early review and so others could begin to use it.

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidLibraryRuleComposer.groovy
cwoodwar6: Certainly.

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidLibraryRuleComposer.groovy
kageiit: If you had this logic in the target, you get access to all source roots easily

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidTestRuleComposer.groovy
```diff
@@ -47,6 +50,7 @@ final class AndroidTestRuleComposer extends AndroidBuckRuleComposer {
                 testAptDeps,
                 providedDeps,
                 aidlRuleNames,
+                extraKotlincArgs,
```
kageiit: Just use target.kotlincArguments directly

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidLibraryRuleComposer.groovy
```diff
@@ -57,6 +60,7 @@ final class AndroidLibraryRuleComposer extends AndroidBuckRuleComposer {
                 libraryAptDeps,
                 providedDeps,
                 aidlRuleNames,
+                extraKotlincArgs,
```
kageiit: Just use target.kotlincArguments directly

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
```diff
@@ -2,13 +2,15 @@ package com.uber.okbuck.core.model.android
...
 import com.android.manifmerger.ManifestMerger2
+import com.uber.okbuck.core.util.KotlinUtil
 import org.gradle.api.Project
-
 /**
  * An Android library target
  */
 class AndroidLibTarget extends AndroidTarget {
 
+    private static final String KOTLIN_EXTENSIONS_OPTION = "plugin:org.jetbrains.kotlin.android:"
```
kageiit: Can we check if the kotlin gradle plugin exposes this constant?

We should try to use it if possible than rolling our own

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
```diff
@@ -22,4 +24,23 @@ class AndroidLibTarget extends AndroidTarget {
     ManifestMerger2.MergeType getMergeType() {
         return ManifestMerger2.MergeType.LIBRARY
     }
+
+    List<String> getKotlincArguments() {
+        List<String> extraKotlincArgs = []
+        if (hasKotlinAndroidExtensions) {
```
kageiit: Lets use an ImmutableList

buildSrc/src/main/groovy/com/uber/okbuck/rule/android/AndroidRule.groovy
```diff
@@ -62,6 +65,14 @@ abstract class AndroidRule extends JavaRule {
             printer.println("\tlanguage = 'kotlin',")
         }
 
+        if (!mExtraKotlincArgs.empty) {
```
kageiit: you can skip the `.empty` in groovy

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
cwoodwar6: I don't see the constant in the kotlin gradle plugin but I do see it in kotlin android extensions itself. But this would require us to add that as a dependency, would you want to do that just for the constant?

kageiit: If it needs a new constant, then No. Hopefully they do not change it. I am not sure they consider this public API or not

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
cwoodwar6: Done

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidLibraryRuleComposer.groovy
cwoodwar6: Done

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidTestRuleComposer.groovy
cwoodwar6: Done

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
```diff
@@ -22,4 +26,23 @@ class AndroidLibTarget extends AndroidTarget {
...
+            getResDirs().each {  String dir ->
+                // :root:module -> /root/module/
+                String module = project.path.replace(":", "/") + "/"
+                // src/main/res -> main
+                String variant = dir.substring(dir.indexOf("/") + 1, dir.lastIndexOf("/"))
+                String pathToRes = project.rootDir.toPath().toAbsolutePath().toString() + module + dir
+                options += KOTLIN_EXTENSIONS_OPTION + "variant=" + variant + ";" + pathToRes + ","
+            }
+            extraKotlincArgs.add("-Xplugin=" + KotlinUtil.KOTLIN_HOME_LOCATION + "/kotlin-android-extensions.jar")
+            extraKotlincArgs.add("-P")
```
kageiit: Are these absolutely paths or relative?

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
```diff
@@ -22,4 +26,23 @@ class AndroidLibTarget extends AndroidTarget {
...
+
+    List<String> getKotlincArguments() {
+        ImmutableList.Builder<String> extraKotlincArgs = ImmutableList.<String>builder()
+        if (hasKotlinAndroidExtensions) {
+            String options = ""
+            getResDirs().each {  String dir ->
+                // :root:module -> /root/module/
+                String module = project.path.replace(":", "/") + "/"
+                // src/main/res -> main
+                String variant = dir.substring(dir.indexOf("/") + 1, dir.lastIndexOf("/"))
```
kageiit: Variant logic is not sound. Any arbitrary directory can be added as a resource directory to a variant.

The variant name should be already available by directly using the `name` field of this target

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
```diff
@@ -22,4 +26,23 @@ class AndroidLibTarget extends AndroidTarget {
...
         return ManifestMerger2.MergeType.LIBRARY
     }
+
+    List<String> getKotlincArguments() {
+        ImmutableList.Builder<String> extraKotlincArgs = ImmutableList.<String>builder()
+        if (hasKotlinAndroidExtensions) {
+            String options = ""
+            getResDirs().each {  String dir ->
+                // :root:module -> /root/module/
+                String module = project.path.replace(":", "/") + "/"
```
kageiit: Use `System.PathSeparator` instead of `/`

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
```diff
@@ -22,4 +26,23 @@ class AndroidLibTarget extends AndroidTarget {
...
+                // src/main/res -> main
+                String variant = dir.substring(dir.indexOf("/") + 1, dir.lastIndexOf("/"))
+                String pathToRes = project.rootDir.toPath().toAbsolutePath().toString() + module + dir
```
kageiit: Path to res is already available via the model which is how `getResDirs()` works. We should use the data from the model directly instead of convoluting it like this

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
cwoodwar6: Absolute

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
cwoodwar6: I think the issue here is that the `name` field will always return a singular `debug` or `main` for each resource directory. I have a case where I need each variant name returned with their appropriate resource path. This way a file in `debug` can reference a resource in `main` and kotlin android extensions appropriately generates for that case.

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
cwoodwar6: 👍

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
cwoodwar6: I need the full absolute path though, is that available and I'm just missing something? If I remember correctly I needed to construct my own path especially for the module name because we have some that are 2 levels deep and that wasn't being accounted for since the project name was only the last of the two directories.

cwoodwar6: I need the full absolute path though, is that available and I'm just missing something? If I remember correctly I needed to construct my own path especially for the module name because we have some that are 2 levels deep and that wasn't being accounted for since the project name was only the last of the two directories.

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
cwoodwar6: For example: My test code uses `import kotlinx.android.synthetic.main.test_layout.view.*` because the path to the resource is `src/main/res`. `name` in that case is always returning `debug`.

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
kageiit: This is still not addressed

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
kageiit: Getting the available resource dirs that are part of the input sourceset is the right way to go then

But I am curious if the variant name really matters? Is there some sort of precedence/ordering being taken into consideration by the kotlin extension here?

How does it work if I have a folder like `custom_resources` added as a resource folder?

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
kageiit: We have to use relative paths or the buck network cache will be rendered ineffective

kageiit:
@cwoodwar6 can you also update .buckversion to latest head and resolve the conflict?

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
cwoodwar6: I added another example that hopefully makes it clear

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
```diff
@@ -22,4 +26,23 @@ class AndroidLibTarget extends AndroidTarget {
     ManifestMerger2.MergeType getMergeType() {
         return ManifestMerger2.MergeType.LIBRARY
     }
+
+    List<String> getKotlincArguments() {
+        ImmutableList.Builder<String> extraKotlincArgs = ImmutableList.<String>builder()
+        if (hasKotlinAndroidExtensions) {
+            String options = ""
```
kageiit: Can we use StringBuilder here for options?

libraries/kotlinandroidlibrary/build.gradle
```diff
@@ -1,8 +1,16 @@
 apply plugin: 'com.android.library'
 apply plugin: 'kotlin-android'
+apply plugin: 'kotlin-android-extensions'
 
 dependencies {
   compile deps.external.kotlinStdlib
   testCompile deps.test.junit
   testCompile deps.test.kotlinTest
 }
+
```
kageiit: The logic in this PR will break if we do something like

```
sourceSets.custom.res.srcDirs = ["src/custom_res"]
```

The above is  valid input to the gradle dsl and all it's tasks, but the way we make assumptions about the folder structure in constructing the extension options is incorrect

cwoodwar6: I updated the custom example to use this

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
```diff
@@ -22,4 +26,23 @@ class AndroidLibTarget extends AndroidTarget {
...
     }
+
+    List<String> getKotlincArguments() {
+        ImmutableList.Builder<String> extraKotlincArgs = ImmutableList.<String>builder()
+        if (hasKotlinAndroidExtensions) {
+            StringBuilder options = new StringBuilder()
+            int index = 0
+            getResDirs().each {  String dir ->
+                // :root:module -> root/module/
+                String module = project.path.replace(":", File.separator).substring(1) + File.separator
```
kageiit: module can be outside the loop. It does not change

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
```diff
@@ -22,4 +26,23 @@ class AndroidLibTarget extends AndroidTarget {
...
+                String variant = getResVariants()[index++]
+                String pathToRes = module + dir
+                options.append(KOTLIN_EXTENSIONS_OPTION + "variant=" + variant + ";" + pathToRes + ",")
```
kageiit: we have StringBuilder so please use append instead of `+` everywhere possible

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
```diff
@@ -22,4 +26,23 @@ class AndroidLibTarget extends AndroidTarget {
     ManifestMerger2.MergeType getMergeType() {
         return ManifestMerger2.MergeType.LIBRARY
     }
+
+    List<String> getKotlincArguments() {
+        ImmutableList.Builder<String> extraKotlincArgs = ImmutableList.<String>builder()
+        if (hasKotlinAndroidExtensions) {
```
kageiit: Check if we have extensions first, otherwise just return `ImmutableList.of()` to reduce number of empty lists we are building

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
```diff
@@ -22,4 +26,23 @@ class AndroidLibTarget extends AndroidTarget {
...
+            StringBuilder options = new StringBuilder()
+            int index = 0
+            getResDirs().each {  String dir ->
+                // :root:module -> root/module/
+                String module = project.path.replace(":", File.separator).substring(1) + File.separator
+                String variant = getResVariants()[index++]
```
kageiit: Can you explain the logic here a bit?

how are we reasoning about the order from `getResDirs()` and `getResVariants()`?

It seems like they belong together in a data object almost?

cwoodwar6: `Set` doesn't guarantee order so this isn't safe to do. I'll instead return the variant name and directory as a Tuple.

cwoodwar6:
Updated to use StringBuilder everywhere and to return directories in a map with their associated variant.

kageiit:
Thanks for cleaning this up!

PR #482, opened 2017-07-29
Overhaul Dependency Caching

kageiit:
This is a major refactor to how dependency caching works in okbuck. Key changes

- Okbuck now resolves every configuration individually and thus does not force using the highest dependency version across all subprojects. **Individual projects can depend on different versions of any dependency now**.
- Classpaths used by buck for build/test are now exactly the same as what gradle uses for its tasks. 
- Moved dependency cache creation logic from a single startup task to individual project `okbuck` tasks. This means operations like downloading dependencies, source jars, extraction of annotation processors, lint jars and proguard configs are all parallelized across subproject tasks.
- Added the concept of a persistent `Store` to cache expensive operations like unzipping aars/jars to extract metadata and walking filetrees to determine source roots. This will help speed up subsequent okbuck runs till we can get fully incremental okbuck tasks.
- Moved dependency validation (usage of snapshots etc.) towards the end of the okbuck lifecycle once all dependencies are materialized. This also makes the error messages more visible.
- Several other improvements to performance/documentation and code readability.


buildSrc/src/main/groovy/com/uber/okbuck/core/dependency/DependencyCache.groovy
```diff
@@ -1,293 +1,260 @@
...
 
-    static boolean isWhiteListed(File depFile) {
-        return WHITELIST_LOCAL_PATTERNS.find { depFile.absolutePath.contains(it) } != null
+    /**
+     * Get the packaged lint jar of an aar dependency if any.
+     *
+     * @param dependency The depenency
+     * @return path to the lint jar in the cache.
+     */
+    String getLintJar(ExternalDependency dependency) {
```
kageiit: lint is part of it because dependencies can have embedded metadata like lint jars. The rest of the lint logic is already in a separate class

buildSrc/src/main/groovy/com/uber/okbuck/core/model/base/Store.java
```diff
@@ -0,0 +1,73 @@
...
+
+    public Store(File storeFile) {
+        this.storeFile = storeFile;
+        props = new Properties();
+        accessed = ConcurrentHashMap.newKeySet();
+
+        if (storeFile.exists()) {
+            try {
+                props.load(new BufferedInputStream(new FileInputStream(storeFile)));
+            } catch (IOException ignored) { }
```
kageiit: There is no danger if it does not load as okbuck will treat it as a fresh build and will just compute metadata again, which is why it is ignored here

buildSrc/src/main/groovy/com/uber/okbuck/core/model/base/Store.java
```diff
@@ -0,0 +1,73 @@
...
+            if (!storeFile.exists()) {
+                storeFile.getParentFile().mkdirs();
+                storeFile.createNewFile();
+            }
+            props.keySet().retainAll(accessed);
+            FileOutputStream out = new FileOutputStream(storeFile);
+            ByteArrayOutputStream arrayOut = new ByteArrayOutputStream();
+            props.store(arrayOut, null);
+            String string = new String(arrayOut.toByteArray(), "8859_1");
+            String sep = System.getProperty("line.separator");
```
kageiit: Could be improved in a separate PR. okbuck is targeting java 8, but I have not had a chance to cleanup all code yet.

buildSrc/src/main/java/com/uber/okbuck/core/dependency/ExternalDependency.java
```diff
@@ -0,0 +1,95 @@
...
+
+import java.io.File;
+
+/**
+ * Represents a pre packaged dependency from an external
+ * source like gradle/maven cache or the filesystem
+ */
+public final class ExternalDependency {
+
+    private static final String LOCAL_DEP_VERSION = "1.0.0";
```
kageiit: This is actually just an arbitrary version number we give to local dependencies i.e deps that are in vcs. It does not mean anything

artem-zinnatullin: Exactly, that's why I'm proposing `-local` to not confuse people :)

buildSrc/src/main/groovy/com/uber/okbuck/core/model/base/Store.java
```diff
@@ -0,0 +1,73 @@
...
+import java.io.IOException;
+import java.util.Properties;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * A persistent data store for simple string key/value pairs
+ * that auto cleans up entries that are no longer accessed.
+ */
+public final class Store {
```
artem-zinnatullin: Would be great to add tests for it

buildSrc/src/main/java/com/uber/okbuck/core/dependency/ExternalDependency.java
kageiit: Good idea. Wanna make a quick PR to fix this?

Otherwise, I will clean it up in a separate commit

buildSrc/src/main/groovy/com/uber/okbuck/core/model/base/Store.java
kageiit: great idea. I will make an issue to keep track

buildSrc/src/main/java/com/uber/okbuck/core/dependency/ExternalDependency.java
artem-zinnatullin: Sure, see https://github.com/uber/okbuck/pull/485

PR #496, opened 2017-08-04
Cleanup some code and use immutables

kageiit:


buildSrc/build.gradle
```diff
@@ -48,7 +48,7 @@ task sourcesJar(type: Jar) {
     classifier = 'sources'
 }
 
-def publishVersion = '0.24.6'
+def publishVersion = '0.24.7-LOCAL'
```
artem-zinnatullin: Seems like something that should be reverted :)

kageiit: haha good catch

PR #513, opened 2017-09-01
Begin migrating buck file generation to template engine

kageiit:
Ref #509

This adds support for writing buck rules based on the https://github.com/fizzed/rocker template engine

Template Comparison

This allows for modifying templates in a more simple way and comes with lot of performance benefits as the templates are now precompiled to java code. This is going to allow incremental execution of okbuck tasks by outputting each rule to an intermediate file for later merging.

### DIFF
--- a/buildSrc/src/main/groovy/com/uber/okbuck/rule/base/GenRule.groovy
+++ b/buildSrc/src/main/groovy/com/uber/okbuck/rule/base/GenRule.groovy
@@ -5,11 +5,7 @@ import com.uber.okbuck.core.model.base.RuleType
 
 final class GenRule extends BuckRule {
 
-    private final Set<String> inputs
-    private final Set<String> bashCmds
-    private final boolean globSrcs
-    private final String output
-    private final boolean executable
+    private final Rule genrule
 
     GenRule(String name,
             List<String> inputs,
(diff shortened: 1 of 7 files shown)


buildSrc/src/main/java/com/uber/okbuck/rule/base/RuleTemplate.java
```diff
@@ -0,0 +1,49 @@
...
+        super.__associate(context);
+
+        if (context instanceof RuleTemplate) {
+            RuleTemplate ninjaContext = (RuleTemplate)context;
+            this.ruleType = ninjaContext.ruleType;
+            this.name = ninjaContext.name;
+            this.visibility = ninjaContext.visibility;
+            this.deps = ninjaContext.deps;
+            this.extraBuckOpts = ninjaContext.extraBuckOpts;
+        }
```
hzsweers: nit: same line the `else`

buildSrc/src/main/groovy/com/uber/okbuck/rule/base/GenRule.groovy
```diff
@@ -18,31 +14,21 @@ final class GenRule extends BuckRule {
             String output = "${name}_out",
             boolean executable = false) {
         super(RuleType.GENRULE, name)
-        this.inputs = inputs
-        this.bashCmds = bashCmds
-        this.globSrcs = globSrcs
-        this.output = output
-        this.executable = executable
+        genrule = new template.base.GenRule()
```
hzsweers: Is it normal for the class to be a nonconventional lowercase `template`?

kageiit: This is just the FQCN as there was a class name conflict

buildSrc/src/main/java/com/uber/okbuck/rule/base/RuleTemplate.java
kageiit: will clean this up in a followup

buildSrc/src/main/groovy/com/uber/okbuck/rule/base/GenRule.groovy
kageiit: The current class will go away and we can drop the FQCN in future PRs

PR #520, opened 2017-09-08
Added integration testing for transform integration 

malbano:
Added integration testing for transform integration with a dummy transform applied to app prod that simply copies all the jars and classes in the output folder.

### DIFF
--- a/dummy-transform/build.gradle
+++ b/dummy-transform/build.gradle
@@ -1,0 +1,9 @@
+plugins {
+    id "groovy"
+}
+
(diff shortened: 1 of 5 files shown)


dummy-transform/build.gradle
```diff
@@ -0,0 +1,9 @@
+plugins {
+    id "groovy"
```
kageiit: just need java

kageiit:
Seems like travis is failing?

kageiit:
You should rebase on master

PR #522, opened 2017-09-09
Write kotlin compiler arg for experimental kotlin android extensions

seanabraham:
When using kotlin android extensions, if you use any experimental features via:
```
androidExtensions {
  experimental = true
}
```
Then the kotlin android extensions gradle plugin adds an extra compiler argument as seen here: https://github.com/JetBrains/kotlin/blob/24c6f5f0f0c6c3d66152caf788c274c9c1832b1e/libraries/tools/kotlin-gradle-plugin/src/main/kotlin/org/jetbrains/kotlin/gradle/internal/AndroidSubplugin.kt#L146

This replicates that logic by writing a similar compiler argument if the AndroidExtensionsExtension is used with `experimental = true` set.


libraries/kotlinandroidexperimentallibrary/src/main/kotlin/com/uber/okbuck/kotlin/KotlinParcelable.kt
```diff
@@ -0,0 +1,7 @@
+package com.uber.okbuck.kotlin
+
+import android.os.Parcelable
+import kotlinx.android.parcel.Parcelize
+
+@Parcelize
+class KotlinParcelable(val name: String) : Parcelable
```
kageiit: just make a new class in the `kotlinlibrary` module. There is no need to make a new one just for this purpose

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidTarget.groovy
```diff
@@ -89,6 +91,15 @@ abstract class AndroidTarget extends JavaLibTarget {
         // Check if kotlin
         isKotlin = project.plugins.hasPlugin(KotlinAndroidPluginWrapper.class)
         hasKotlinAndroidExtensions = project.plugins.hasPlugin(KOTLIN_ANDROID_EXTENSIONS_MODULE)
+        try {
```
kageiit: You do not need to try if `hasKotlinAndroidExtensions` is true as the extension is guaranteed to be available

libraries/kotlinandroidexperimentallibrary/src/main/kotlin/com/uber/okbuck/kotlin/KotlinParcelable.kt
seanabraham: I made the new library to verify that the compiler arguments were indeed different with experimental=true specified. Given that, still think I should just make this a class in the existing `kotlinandroidlibrary`?

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidTarget.groovy
seanabraham: 👍

libraries/kotlinandroidexperimentallibrary/src/main/kotlin/com/uber/okbuck/kotlin/KotlinParcelable.kt
kageiit: Ya I think having a separate module just for one extension option is not scalable. As we add more extension options, it will get harder to maintain down the line

artem-zinnatullin: Isn't that the way Buck wants you to structure projects :trollface:

seanabraham: Done

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidTarget.groovy
```diff
@@ -24,6 +24,7 @@ import org.gradle.api.JavaVersion
 import org.gradle.api.Project
 import org.gradle.api.tasks.compile.JavaCompile
 import org.gradle.api.tasks.testing.Test
+import org.jetbrains.kotlin.gradle.internal.AndroidExtensionsExtension
```
kageiit: I am a little worried about relying on an internal class here. The `getByType` api uses reflection under the hood anyways. Can we please just do this instead?

- Remove the extension dependency in build.gradle
- `hasExperimentalKotlinAndroidExtensions = project.androidExtensions.experimental` in a try catch block guarded by `hasKotlinAndroidExtensions`

seanabraham: Sure

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidTarget.groovy
```diff
@@ -90,6 +91,13 @@ abstract class AndroidTarget extends JavaLibTarget {
         isKotlin = project.plugins.hasPlugin(KotlinAndroidPluginWrapper.class)
         hasKotlinAndroidExtensions = project.plugins.hasPlugin(KOTLIN_ANDROID_EXTENSIONS_MODULE)
 
+        try {
+            hasExperimentalKotlinAndroidExtensions = hasKotlinAndroidExtensions &&
+                                                     project.androidExtensions.experimental
+        } catch (Exception ignored) {
```
cwoodwar6: Do you still need the catch block?

seanabraham: We can probably get rid of the try-catch all together but I'm not sure if there are cases that I'm not testing for where `hasKotlinAndroidExtension` is true but `project.androidExtensions` does not exist

seanabraham:
🎉 , thanks @kageiit!

PR #524, opened 2017-09-19
Pass the merged manifest to Robolectric tests for app modules

jbarr21:
This is useful to be able to write tests that catch all activities or permissions that run only once per app. This passes the merged Android manifest for app modules but not for library modules.

### DIFF
--- a/buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidTestRuleComposer.groovy
+++ b/buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidTestRuleComposer.groovy
@@ -1,6 +1,7 @@
 package com.uber.okbuck.composer.android
 
 import com.google.common.collect.ImmutableSet
+import com.uber.okbuck.core.model.android.AndroidAppTarget
 import com.uber.okbuck.core.model.android.AndroidLibTarget
 import com.uber.okbuck.core.model.base.RuleType
 import com.uber.okbuck.core.util.RetrolambdaUtil
@@ -20,7 +21,8 @@ final class AndroidTestRuleComposer extends AndroidBuckRuleComposer {
             AndroidLibTarget target,
             List<String> deps,
             final List<String> aidlRuleNames,
-            String appClass) {
+            String appClass,
+            String manifestRuleName) {
 
         List<String> testDeps = new ArrayList<>(deps)
         testDeps.add(":${src(target)}")
@@ -55,7 +57,7 @@ final class AndroidTestRuleComposer extends AndroidBuckRuleComposer {
                 .options(target.main.jvmArgs)
                 .jvmArgs(target.testOptions.jvmArgs)
                 .env(target.testOptions.env)
-                .robolectricManifest(fileRule(target.manifest))
+                .robolectricManifest(manifestRuleName != null ? manifestRuleName : fileRule(target.manifest))
                 .runtimeDependency(RobolectricUtil.ROBOLECTRIC_CACHE)
 
         if (target.testRuleType == RuleType.KOTLIN_ROBOLECTRIC_TEST) {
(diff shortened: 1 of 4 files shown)


buildSrc/src/main/groovy/com/uber/okbuck/generator/BuckFileGenerator.groovy
```diff
@@ -180,11 +180,19 @@ final class BuckFileGenerator {
 
         // Test
         if (target.robolectric && target.test.sources && !target.isTest) {
+            String manifestRuleName = null
+            if (target instanceof AndroidAppTarget) {
+                Rule manifestRule = AndroidManifestRuleComposer.compose(target)
```
kageiit: this is quite heavy.

If we just need the name, its literally just `manifest_${target.name}`

Use the `manifest` static method on any target to get this. There is no need to make another rule

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidTestRuleComposer.groovy
```diff
@@ -55,7 +57,7 @@ final class AndroidTestRuleComposer extends AndroidBuckRuleComposer {
                 .options(target.main.jvmArgs)
                 .jvmArgs(target.testOptions.jvmArgs)
                 .env(target.testOptions.env)
-                .robolectricManifest(fileRule(target.manifest))
+                .robolectricManifest(manifestRuleName != null ? manifestRuleName : fileRule(target.manifest))
```
kageiit: lets just use `manifest(target)` here directly

We should remove `manifestRuleName`

buildSrc/src/main/groovy/com/uber/okbuck/generator/BuckFileGenerator.groovy
```diff
@@ -180,11 +180,14 @@ final class BuckFileGenerator {
 
         // Test
         if (target.robolectric && target.test.sources && !target.isTest) {
+            String manifestRuleName = target instanceof AndroidAppTarget ? ":${AndroidBuckRuleComposer.manifest(target)}" : null
```
kageiit: dont need this anymore

buildSrc/src/main/groovy/com/uber/okbuck/generator/BuckFileGenerator.groovy
```diff
@@ -180,11 +180,14 @@ final class BuckFileGenerator {
...
+
             androidLibRules.add(AndroidTestRuleComposer.compose(
                     target,
                     deps,
                     aidlRuleNames,
-                    appClass))
+                    appClass,
+                    manifestRuleName))
```
kageiit: dont need this anymore

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidTestRuleComposer.groovy
```diff
@@ -40,6 +41,8 @@ final class AndroidTestRuleComposer extends AndroidBuckRuleComposer {
             providedDeps.add(RetrolambdaUtil.getRtStubJarRule())
         }
 
+        String manifest = target instanceof AndroidAppTarget ? ":${AndroidBuckRuleComposer.manifest(target)}" : fileRule(target.manifest)
```
kageiit: Why the if/else? Android lib targets can merge manifests too. We should use the merged manifest always

jbarr21: I did this for perf savings since the case for using a merged manifest in a lib module test is less convincing. Happy to change for consistency though

hzsweers:
Let's make sure that gets into the wiki here somewhere because it seems preeeeetty important

kageiit:
> Let's make sure that gets into the wiki here somewhere because it seems preeeeetty important

Not really. This is very specific to an internal usecase that we can solve with defs. This makes the builds slower so it should not be a default option as it is not needed by robolectric to run tests

hzsweers:
Merged manifests isn't normal?

kageiit:
Not sure what you mean by normal. It's just not required.

PR #533, opened 2017-10-10
Fix OkBuck exception when buildscript dependency is Jar

felipecsl:
When using Gradle buildscript dependencies as jar files, OkBuck would throw a 
```
Caused by: java.lang.ClassCastException: org.gradle.internal.component.local.model.OpaqueComponentArtifactIdentifier cannot be cast to org.gradle.api.artifacts.component.ModuleComponentIdentifier
        at com.uber.okbuck.core.util.ProjectUtil.lambda$findVersionInClasspath$0(ProjectUtil.java:83)
        at com.uber.okbuck.core.util.ProjectUtil.findVersionInClasspath(ProjectUtil.java:87)
        at com.uber.okbuck.core.util.LintUtil.getDefaultLintVersion(LintUtil.java:31)
        at com.uber.okbuck.extension.LintExtension.<init>(LintExtension.java:28)
        at com.uber.okbuck.extension.LintExtension_Decorated.<init>(Unknown Source)
        at org.gradle.internal.reflect.DirectInstantiator.newInstance(DirectInstantiator.java:51)
        ... 134 more
```

This fixes it by filtering the stream to only `ComponentIdentifier` instances that are `ModuleComponentIdentifier`.

Example test case:
```
buildscript {
  dependencies {
    classpath files("libs/foobar.jar")
  }
}
```

It wasn't obvious where to write a test for this, but I'd be happy to write one if you could provide pointers.

### DIFF
--- a/buildSrc/src/main/java/com/uber/okbuck/core/util/ProjectUtil.java
+++ b/buildSrc/src/main/java/com/uber/okbuck/core/util/ProjectUtil.java
@@ -1,5 +1,8 @@
 package com.uber.okbuck.core.util;
 
+import java.io.File;
+import java.util.Map;
+
 import com.android.build.gradle.AppPlugin;
 import com.android.build.gradle.LibraryPlugin;
 import com.uber.okbuck.OkBuckGradlePlugin;
@@ -8,7 +11,6 @@ import com.uber.okbuck.core.model.base.ProjectType;
 import com.uber.okbuck.core.model.base.Scope;
 import com.uber.okbuck.core.model.base.Target;
 import com.uber.okbuck.core.model.base.TargetCache;
-
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.component.ModuleComponentIdentifier;
 import org.gradle.api.plugins.GroovyPlugin;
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/java/com/uber/okbuck/core/util/ProjectUtil.java
```diff
@@ -78,14 +77,13 @@ static String findVersionInClasspath(Project project, String group, String modul
                 .getArtifacts()
                 .getArtifacts()
                 .stream()
-                .filter(resolvedArtifactResult -> {
-                    ModuleComponentIdentifier identifier =
-                            (ModuleComponentIdentifier) resolvedArtifactResult.getId().getComponentIdentifier();
-                    return (group.equals(identifier.getGroup()) &&
-                            module.equals(identifier.getModule()));
-                })
+                .map(artifactResult -> artifactResult.getId().getComponentIdentifier())
```
kageiit: can you reduce the amount of nested map/filters here? maybe flat map it?

felipecsl:
Done, PTAL

PR #541, opened 2017-10-19
Do not use merged manifest for lint and unit tests

kageiit:
Looking at the source of the android gradle plugin, it seems like the merged manifest is not used for lint or for running unit tests. Mimic'ing the same behavior


artem-zinnatullin:
Hm, how would permission Lint rule(s) work then…? Library may add some permissions to the app

kageiit:
@hzsweers can confirm about lint behavior as he dug into the source

PR #561, opened 2017-11-24
Update buckw

ankit-mnnit:
If this fix is applied, then buckw will be able to handle directory names(path name) like ~/abc - tue/

The script was stuck when such names were found in path.

### DIFF
--- a/buckw
+++ b/buckw
@@ -71,7 +71,7 @@ OKBUCK_DIR="$SCRIPT_DIR/.okbuck"
 MAX_DISPLAY_CHANGES=10
 
 ensureWatch ( ) {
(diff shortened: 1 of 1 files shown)


kageiit:
Can you change it in https://github.com/uber/okbuck/blob/master/buildSrc/src/main/rocker/com/uber/okbuck/template/config/BuckWrapper.rocker.raw as well? That is the template okbuck uses to generate buckw for consumers

kageiit:
@ankit-mnnit any update on the comment?

kageiit:
5332c0945ac8fa45763603f938155638007dbc54

PR #575, opened 2018-01-15
Add better error messages for multiple proguard files

runningcode:
If more than one proguard file is detected, fail the build because
okbuck will not use the other files

### DIFF
--- a/buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidAppTarget.groovy
+++ b/buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidAppTarget.groovy
@@ -3,6 +3,7 @@ package com.uber.okbuck.core.model.android
 import com.android.build.gradle.api.BaseVariant
 import com.android.builder.model.SigningConfig
 import com.android.manifmerger.ManifestMerger2
+import com.google.common.base.Preconditions
 import com.google.common.collect.ImmutableList
 import com.google.common.collect.ImmutableSet
 import com.uber.okbuck.core.util.FileUtil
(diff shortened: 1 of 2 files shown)


buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidAppTarget.groovy
```diff
@@ -113,27 +115,27 @@ class AndroidAppTarget extends AndroidLibTarget {
         if (minifyEnabled) {
             Set<File> proguardFiles = (baseVariant.mergedFlavor.proguardFiles +
                     baseVariant.buildType.proguardFiles)
-            if (proguardFiles.size() > 0 && proguardFiles[0].exists()) {
-                File genProguardConfig = getGenPath("proguard.pro")
+            Preconditions.checkArgument(proguardFiles.size() == 1, "%s proguard files found. Only one can be used.", proguardFiles.size())
```
kageiit: do you mean `proguardFiles.size() != 1`

kageiit:
also, why is exopackage related to multidex?

i think the bug you are trying to fix should be if exo is enabled, always write out `should_split_dex = True` in the buck file. Having it be set by turning multidex support on is a weird side effect. I think we should just fix the exopackage split dex flag logic instead

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidAppTarget.groovy
runningcode: it throws an exception if the condition is not true. in this case we must have at least one proguard file since `minifyEnabled == true` but we cannot have more than one because okbuck will ignore it.

runningcode:
Ok, I removed the multidex exception. This will only throw exception for the proguard misconfigurations. I'll file an upstream bug to buck about the multidex/exopackage issue.

PR #579, opened 2018-01-25
Switch okbuck to use prebuilt buck binaries

kageiit:
- Enables okbuck to download prebuilt buck binaries from jitpack added as part of https://github.com/facebook/buck/pull/1712 . This means okbuck will no longer build buck locally on the machine and will also not require `ant` to be installed. This should lead to faster updates and quick switching between buck versions.
- Version of buck to fetch can be customized via the okbuck extension like
```
okbuck {
  buckBinary = "com.github.facebook:buck:26845eb89ba4464586787533d08984a993f247d4@pex"
}
```
- Okbuck will automatically add jitpack as a remote repository on the root project to fetch prebuilt buck binaries. The very first time a version is pulled, it can take a few minutes to build buck on jitpack, so you may experience timeouts fetching the binary. Just retry and from every next time, the artifact would be cached and served by jitpack as usual.
- To download custom buck from a fork, if the fork is on github, you can change the maven coords appropriately. For example, for a fork at https://github.com/kageiit/buck one may use `"com.github.kageiit:buck:26845eb89ba4464586787533d08984a993f247d4@pex"`
- If your fork is not on github, you can still configure okbuck to download prebuilt buck binaries from your maven server like so in your root `build.gradle` (you will have to manage uploading binaries to your maven server via your own infra as repos not on github are not visible to jitpack)
```
okbuck {
  buckBinary = null
}

repositories {
  maven { url <your-custom-maven-or-artifactory-url> }
}

dependencies {
  buckBinary "com.example:buck:26845eb89ba4464586787533d08984a993f247d4@pex"
}
```


kageiit:
cc @cwoodwar6 @seanabraham @artem-zinnatullin @runningcode

runningcode:
This is nice! Not having the ant dependency makes it a bit nicer to use on CI as well.
Maybe worth updating this wiki page once merged: https://github.com/uber/okbuck/wiki/OkBuck-options-and-examples

kageiit:
> Maybe worth updating this wiki page once merged: https://github.com/uber/okbuck/wiki/OkBuck-options-and-examples

Yep plan to update both wiki and readme

buckw
```diff
@@ -59,18 +59,14 @@ jsonq() {
 }
 
 ensure python 'https://www.python.org'
-ensure ant 'http://ant.apache.org/'
 INSTALLED_WATCHMAN=`command -v watchman`
 
-DEFAULT_BUCK_REPO="https://github.com/facebook/buck.git"
-DEFAULT_BUCK_INSTALL_DIR="$HOME/.gradle/caches/okbuck/buck"
```
cwoodwar6: Do we want to provide some sort of cleanup? This leaves the old repo checked out.

kageiit: I thought about it, but I did not want to force folks to reclone the buck repo if they were building an older sha for some reason.

That should be upto the users (who can modify their buckw to do the cleanup if they like)

PR #586, opened 2018-02-19
Fix watchman config default ignores.

runningcode:
The comma at the end was causing watchman to silently ignore the config.


kageiit:
Can you update the `.watchmanconfig` at root also?

PR #589, opened 2018-02-20
Fix buckBinary configuration name: _ to CamelCase.

artem-zinnatullin:
Was trying to figure out why am I getting this error when I'm trying to use custom Buck binary as recommended in [changelog for 0.32.2](https://github.com/uber/okbuck/releases/tag/v0.32.2):

```console
Could not find method buckBinary() for arguments [our.custom.buck:buck:someversion@pex] on object of type org.gradle.api.internal.artifacts.dsl.dependencies.DefaultDependencyHandler.
```

Turned out changelog says to use `buckBinary` configuration while actual configuration name is `buck_binary` :trollface: 

Alternative fix would be to update changelog, but Gradle typically uses CamelCase for configuration names.

### DIFF
--- a/buildSrc/src/main/groovy/com/uber/okbuck/OkBuckGradlePlugin.groovy
+++ b/buildSrc/src/main/groovy/com/uber/okbuck/OkBuckGradlePlugin.groovy
@@ -63,7 +63,8 @@ class OkBuckGradlePlugin implements Plugin<Project> {
     public static final String TRANSFORM = "transform"
     public static final String SCALA = "scala"
     public static final String FORCED_OKBUCK = "forcedOkbuck"
-    public static final String BUCK_BINARY = "buckBinary"
+    public static final String BUCK_BINARY = "buck_binary"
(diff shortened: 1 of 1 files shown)


kageiit:
You will need 2 constants. One for the configuration and one for the cache folder. You may want to leave `BUCK_BINARY` alone and make a new one `BUCK_BINARY_CONFIGURATION` for the configuration bit

artem-zinnatullin:
Yeah I've noticed that cache folder will be different with this patch, do you think that's a big deal?

Though I'm curious why do we have to copy Buck binary from `~/.gradle/caches/modules-*/files-*/…` to project `.okbuck/cache` and not just work with the one in Gradle cache?

kageiit:
Lets keep the cache folder names the same. We dont use camel case in the cache folder.

It is not strictly necessary to copy the files over. we dont do so internally, but refactoring the wrapper to look in gradle caches in going to come up in a follow up change. Lets keep the folder names intact for now.

artem-zinnatullin:
Sounds good!

artem-zinnatullin:
Updated

PR #591, opened 2018-02-26
Add apt api/implementation problem sample.

artem-zinnatullin:
This is a problem we're facing with `api`/`implementation` migration and annotation processing.

Looks like `implementation` hides classes from annotation processors too even though from public methods standpoint we're keeping things private/package-private. 

#### Gradle

```console
./gradlew :libraries:dagger1:module1:build
```

Succeeds (!) with following log:

```console
> Task :libraries:dagger1:module1:compileJava 
error: cannot access Class3
  class file for com.module3.Class3 not found
  Consult the following stack trace for details.
  com.sun.tools.javac.code.Symbol$CompletionFailure: class file for com.module3.Class3 not found


BUILD SUCCESSFUL in 1s
```

You can notice that actually there is a error in the log, but required classes gets generated and build is successful. 

#### Buck

```console
./buckw build //libraries/dagger1/module1:src_main
```

Fails with following log:

```console
Build failed: 
+------------------------------------------------------------------------------------------------+
|                                                                                                |
| The annotation processor dagger.internal.codegen.GraphAnalysisProcessor has crashed.           |
|                                                                                                |
| This is likely a bug in the annotation processor itself, though there may be changes you can   |
| make to your code to work around it. Examine the exception stack trace below and consult the   |
| annotation processor's troubleshooting guide.                                                  |
|                                                                                                |
| com.sun.tools.javac.code.Symbol$CompletionFailure: class file for com.module3.Class3 not found |
|                                                                                                |
+------------------------------------------------------------------------------------------------+

    When running javac
    When running <javac_jar>.
    When building rule //libraries/dagger1/module1:src_main.
```

---

Replacing `implementation` with `api` in `libraries/dagger1/module2/build.gradle` "fixes" build with Buck and removes error from Gradle build log.

Should we propagate all transitive dependencies to annotation processors no matter if they're `implementation` or not?

 @tbroyer maybe you have insights on why Gradle notices the error but succeeds the build and is it even valid behavior?

---

P.S. Example uses Dagger1 because Dagger2 seems to do `@Module(library = true, complete = false)` behavior automatically while with Dagger1 we can specify only `@Module(library = true)` without `complete = false` which allows us reproduce the problem.


tbroyer:
> @tbroyer maybe you have insights on why Gradle notices the error but succeeds the build and is it even valid behavior?

Running the build with `--debug` gives you the arguments passed to the Java compiler. Running the compiler from the command line with the same arguments produces the same result: message printed to stderr but build successful:
```
$ javac -source 1.8 -target 1.8 -d /tmp/okbuck/libraries/dagger1/module1/build/classes/java/main -s /tmp/okbuck/libraries/dagger1/module1/build/generated/source/apt/main -g -sourcepath ""  -processorpath /home/tbr/.gradle/caches/modules-2/files-2.1/com.squareup.dagger/dagger-compiler/1.2.5/6acfbfe76cb96dcf88144845844eeca7314fc0d3/dagger-compiler-1.2.5.jar:/home/tbr/.gradle/caches/modules-2/files-2.1/com.squareup.dagger/dagger/1.2.5/965d3e29792e7c4193a2906ccadd5ec2fbb6d441/dagger-1.2.5.jar:/home/tbr/.gradle/caches/modules-2/files-2.1/com.squareup/javapoet/1.7.0/4fdcf1fc27c1a8f55d1109df986c923152f07759/javapoet-1.7.0.jar:/home/tbr/.gradle/caches/modules-2/files-2.1/com.google.guava/guava/19.0/6ce200f6b23222af3d8abb6b6459e6c44f4bb0e9/guava-19.0.jar:/home/tbr/.gradle/caches/modules-2/files-2.1/javax.inject/javax.inject/1/6975da39a7040257bd51d21a231b76c915872d38/javax.inject-1.jar -XDuseUnsharedTable=true -classpath /tmp/okbuck/libraries/dagger1/module2/build/classes/java/main:/home/tbr/.gradle/caches/modules-2/files-2.1/com.squareup.dagger/dagger/1.2.5/965d3e29792e7c4193a2906ccadd5ec2fbb6d441/dagger-1.2.5.jar:/home/tbr/.gradle/caches/modules-2/files-2.1/javax.inject/javax.inject/1/6975da39a7040257bd51d21a231b76c915872d38/javax.inject-1.jar /tmp/okbuck/libraries/dagger1/module1/src/main/java/com/module1/DaggerModule1.java
error: cannot access Class3
  class file for com.module3.Class3 not found
  Consult the following stack trace for details.
  com.sun.tools.javac.code.Symbol$CompletionFailure: class file for com.module3.Class3 not found
$ echo $?
0
```

That being said, `Class3` is part of the API of `DaggerModule2`, which you use from `module1`, so `module3` needs to be exposed to `module1` transitively, i.e. should be declared as `api` in `module2`, not `implementation`.

artem-zinnatullin:
Unless I'm missing something it seems to me that we should provide all transitive dependencies to annotation processors no matter if they were hidden with `implementation` or not because there are complex annotation processors like Dagger1/2 which might need to cross module visibility boundaries in order to discover all the types required for annotation processing

artem-zinnatullin:
Although, Gradle docs mention:

>types used in public method parameters, including generic parameter types (where public is something that is visible to compilers. I.e. , public, protected and package private members in the Java world)

In [section about API/ABI](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_recognizing_dependencies)

In this particular example we can't declare visibility of `@Provides` method lower than package-private, but it's more a limitation of Dagger1/2 rather than general apt problem. 
I can try to build another annotation processor that will need to cross module visibility barriers while keeping annotated methods `private`

tbroyer:
> Unless I'm missing something it seems to me that we should provide all transitive dependencies to annotation processors no matter if they were hidden with implementation or not because there are complex annotation processors like Dagger1/2 which might need to cross module visibility boundaries in order to discover all the types required for annotation processing

I can't speak for okbuck, but "provid[ing] all transitive dependencies to annotation processors" is just about putting them in the compilation classpath, which is exactly what `api` does vs. `implementation`. But those _complex annotation processors_ are identified and won't reach into any class they're not told to reach into; i.e. here, Dagger triggers the loading of `Class3` because it processes the `DaggerModule2` (by way of `DaggerModule1` referencing it). `module2` explicitly _exposes_ `Class3` to Dagger in downstream modules, so it should declare `module3` as `api` **or** "tell" downstream modules to add an explicit dependency on `module3` if they use the `DaggerModule2` module. The same would be true if `Class2` had a dependency on `Class3` and used an `@Inject` constructor rather than a `@Provides` method in `DaggerModule2`.

artem-zinnatullin:
I've refactored example and replaced `Class2` with `Interface2` to emphasize that `Class3` is hidden in terms of API of the `module2`, this is how we actually use Dagger in our project.

As you can see we now expose `Interface2` which doesn't say a thing about `Class3`, while implementation of `Interface2` uses `Class3` and that's a big point of having interfaces — hide implementation details like this.

However we're required to expose `module3` just for proper annotation processing which defeats the purpose of `api`/`implementation` :(

Hope that makes sense now

tbroyer:
It **did** make sense already. What I'm saying is that `DaggerModule2` **is** part of your API too, and brings `Class3` with it as a transitive dependency.

The alternative would be to declare another `implementation project(":module3")` dependency in `module1`, as it uses `DaggerModule2` which is the only API of `module2` that actually _exposes_ `module3` (what I meant by "tell downstream modules to add an explicit dependency on module3 if they use the DaggerModule2 module").
(or move `DaggerModule2` to another module so it can expose `module3` without `module2` itself exposing `module3` as a transitive dependency; but that's only useful if you actually use `module2` without referencing `DaggerModule2`).
…or you could really _encapsulate_ things by having `DaggerModule2` delegating to another `ObjectGraph` and only effectively exposing `Interface2` (similar to subcomponents vs. component dependencies in Dagger 2)

artem-zinnatullin:
`DaggerModule2` is for sure part of `module2` API, however it doesn't really expose `Class3` except package-private declaration which I'm very sure could have been be private if Dagger would support private `@Provide` methods and we would still have exact same cross-module visibility problems

All solutions you've suggested **will work** technically, but in a way defeat benefits of using `api`/`implementation` :(

---

I've added Guice example that has same module composition but since it's a runtime DI it's not affected by this problem and both Buck and Gradle compile and run it just fine and we're able to use benefits of `implementation` configuration.

Not sure I'm proving something with Guice example other than well, yeah that runtime based DI allows us defer `Class3` lookup and keep `module3` hidden from `module1`

I still feel that extending scope of apt dependencies is a better way since it only leaks it for apt step and it can be done automatically, while playing with `api`/`implementation` has drawbacks in both scenarios:

- If `module2` declares `module3` as `api` that exposes `module3` to all users of `module2` no matter if they do apt or not and affect build times as well as desired module visibility
- If `module2` declares `module3` as `implementation` — all users of `module2` will have to declare `module3` as `implementation` as well which is super repetitive manual work that looks like a workaround to make apt work

Separating Gradle modules from Dagger modules and recomposing them looks like a big refactoring and makes things unnecessary harder if you use DI as foundation of plugin system for your feature modules.

So far all of these solutions other than extended apt deps scope look pretty problematic to implement (specifically in our project with 437 modules as of today) :(

We've tried so hard to hide implementation details by separating interfaces and implementations and now this thing breaks whole idea 😿 

Maybe this could be added as an option to your apt plugin and then OkBuck could also read it and replicate for Buck?

kageiit:
@artem-zinnatullin Gradle 4.6 nd above come with the `annotationProcessor` configuration (this is also what AGP uses). Okbuck just fetches dependencies that AGP exposes. Adding more deps on top of what the standard configurations expose is not something okbuck should be doing.

I suggest you check with gradle 4.6 (rc2 is out) and see what the behavior is for java projects

artem-zinnatullin:
@kageiit for sure I don't want OkBuck behavior be different, that's why I cced @tbroyer as we rely on his apt plugin

However even right now Buck and Gradle behavior is different and we might want to patch Buck to align them since as @tbroyer shown `javac` on its own doesn't fail but prints error, while Buck also fails

kageiit:
> we might want to patch Buck to align them

Can you open an upstream issue on buck? I am not sure if its an issue in gradle or buck

kageiit:
@artem-zinnatullin According to gradle 4.6, https://github.com/gradle/gradle/blob/95dcf58866cc0065d89d500562b85d3a2496bff2/subprojects/plugins/src/main/java/org/gradle/api/plugins/JavaBasePlugin.java#L284

The `annotationProcessor` configuration does not extend from any other configurations and is fully isolated. The issues you pointed out are specific to Dagger trying to resolve the classes it needs to for checking injection and providers match up at compile time (as you noted, a runtime based DI like guice does not have this problem). I think this is a problem if you rely on this as part of your app design as the only option to have compile time verification via dagger is to make the class symbol available to it.

As another observation,
```
If module2 declares module3 as api that exposes module3 to all users of module2 no matter if they do apt or not and affect build times as well as desired module visibility
```
This statement is not correct if you use [depfiles](https://buckbuild.com/concept/buckconfig.html#build.depfiles) in buck which is on by default. If a consumer that gets module 2 exposed to it actually never refers to any code from it, it will not get recompiled even if you change public api in module 2. This is a big advantage of the buck build process compared to gradle so I think making `module2` an api dep is actually the simplest solution and will not affect build times unnecessarily.

tbroyer:
What changes things here is the use of an annotation processor (Dagger) that needs to look at classes from the dependency project. You could probably also use a project dependency specifying a specific configuration, and creating such a configuration exposing the appropriate dependencies (I'm not very familiar with this, so the snippet below might not work, or not as expected):

```gradle
configurations {
  dagger {
    extendsFrom apiElements
  }
  // declare implementation to extend from dagger so you don't have to declare those dependencies twice
  implementation.extendsFrom dagger
}
dependencies {
  dagger project(":module3") { transitive = false } // because you don't need transitive dependencies in this specific case
  implementation project(":module3") // override to get transitive dependencies
}
```
and then in module1:
```gradle
dependencies {
  dagger project(":module2", "dagger")
}
```

I have no idea how OkBuck would turn this into BUCK files though, so might not work there too (even if it worked in Gradle proper).

---

> All solutions you've suggested **will work** technically, but in a way defeat benefits of using `api`/`implementation` :(

I suppose it depends what the "benefits of `api`/`implementation`" are for you.
I'm sure Gradle Inc. could help you find a solution that'd still play well with build times.

> I still feel that extending scope of apt dependencies is a better way since it only leaks it for apt step

The thing is: there's no "apt step", it's the compilation step. If you want/need an annotation processor to see a class, it has to be on the compilation classpath; this in itself defeats the "module visibility" part of `api`/`implementation`. You could technically run the compilation once with `-proc:only` and `module3` in the classpath, then once with `-proc:none`, using the generated sources as additional inputs to be compiled, and without `module3` in the classpath; that still wouldn't solve the "build time" issue.
For that, you need real segregation of the `module3` API, with `DaggerModule2` somehow never exposing `module3` classes (in this case, it could delegate to another `ObjectGraph`, but that would stop working as soon as you introduce singletons in the equation).

I don't have a solution, other than not using Dagger *or* living with the consequences.

> Maybe this could be added as an option to your apt plugin and then OkBuck could also read it and replicate for Buck?

No. Because the goal of the plugin always was to eventually be integrated upstream and go away. And this is happening: Gradle 3.4 introduced CompileOptions.annotationProcessorPath, Gradle 4.3 added CompileOptions.annotationProcessorGeneratedSourcesDirectory, and Gradle 4.6 introduces annotationProcessor configurations and SourceSet.annotationProcessorPath. The plugin, in Gradle 4.6, finally only adds a default value for `CompileOptions.annotationProcessorGeneratedSourcesDirectory`, and a DSL to more easily set processor args and/or the list of processors to run (which I'm probably going to deprecate at some point). Better IDE integration (net.ltgt.apt-eclipse and net.ltgt.apt-idea plugins) will be added to Gradle proper at some point too, and hopefully to Buildship and IntelliJ IDEA.

kageiit:
@artem-zinnatullin going to close for inactivity. I think there have been several good workarounds suggested already for your issue

PR #615, opened 2018-04-05
 Migrate Scope.groovy to java

raviagarwal7:
Verified the buck files remained the same before and after the change.


buildSrc/src/main/java/com/uber/okbuck/extension/OkBuckExtension.java
```diff
@@ -138,4 +139,14 @@
...
         buckProjects = project.getSubprojects();
     }
+
+    private IntellijExtension intellijExtension;
+
+    public IntellijExtension getIntellijExtension() {
+        return intellijExtension;
+    }
+
+    public void setIntellijExtension(IntellijExtension extension) {
```
runningcode: what does this look like when you set this from the groovy DSL?

i was thinking it might be easier to use this extension if we have a method like
```
public void intellij(Action<IntellijExtension> container) {
  container.execute(new IntellijExtension());
}
```
thoughts?

I looked at the source for BaseExtension in the android gradle plugin. the SpotlessExtension in this project uses a similar pattern:
https://github.com/diffplug/spotless/blob/master/plugin-gradle/src/main/java/com/diffplug/gradle/spotless/SpotlessExtension.java

raviagarwal7: 👍 good idea, using this format now since it doesn't require you to explicitly set it.

PR #616, opened 2018-04-06
Android libraries instrumentation apks

thalescm:
This PR adds the functionality to create build rules for android libraries instrumentation apks.

## Notes

* A new rule was added, `RobolectricRule` which extends from `AndroidRule`.
This way we can filter it and not add it as a dependency on instrumentation apks src rules. This was happening not only in library instrumentation apks, but app ones too, so now we don't need to compile unit test src code to compile ui tests 🎉 
* UI tests resources rules didn't depend on the non test resources rule. Added this functionality too (for both app and lib instrumentation tests)


thalescm:
I will (it's on the PR description of missing things).

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidBuckRuleComposer.groovy
```diff
@@ -34,7 +34,7 @@ abstract class AndroidBuckRuleComposer extends JvmBuckRuleComposer {
         return "app_lib_${target.name}"
     }
 
-    static String bin(AndroidAppTarget target) {
```
kageiit: Having `bin_` for android libraries does not make sense. Lets name these differently since the `bin_` is supposed to be for actual apps.

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibInstrumentationTarget.groovy
```diff
@@ -0,0 +1,40 @@
+package com.uber.okbuck.core.model.android
+
+import com.android.build.gradle.api.BaseVariant
+import com.android.build.gradle.api.LibraryVariant
+import com.android.builder.model.SourceProvider
+import org.gradle.api.Project
+/**
```
kageiit: newline after last import

buildSrc/src/main/groovy/com/uber/okbuck/generator/BuckFileGenerator.groovy
```diff
@@ -222,21 +219,47 @@ final class BuckFileGenerator {
...
 
         rules.add(AndroidInstrumentationApkRuleComposer.compose(filterAndroidDepRules(rules), target, mainApkTarget))
         rules.add(AndroidInstrumentationTestRuleComposer.compose(mainApkTarget))
         return rules
     }
 
+    private static List<Rule> createRules(AndroidLibInstrumentationTarget target, List<Rule> mainLibTargetRules) {
+        List<Rule> rules = []
+
+        // TODO: We should find a way to filter robolectric rule too,
```
kageiit: Remove the TODO?

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibTarget.groovy
```diff
@@ -14,8 +17,18 @@ class AndroidLibTarget extends AndroidTarget {
 
     private static final String KOTLIN_EXTENSIONS_OPTION = "plugin:org.jetbrains.kotlin.android:"
 
+    final AndroidLibInstrumentationTarget libInstrumentationTarget
+
     AndroidLibTarget(Project project, String name, boolean isTest = false) {
         super(project, name, isTest)
+
+        TestExtension testExtension = rootProject.okbuck.test
+        if (testExtension.espresso && instrumentationTestVariant && ProjectUtil.getType(project) == ProjectType.ANDROID_LIB) { // do not try to create this for android apps
```
kageiit: This adds a lot of performance overhead for projects with a lot of library projects and no instrumentation tests on libraries.

I would recommend gating this entire functionality behind an okbuck extension flag like

```
okbuck {
  test {
    espressoForLibraries = true
  }
}
```

libraries/kotlinandroidlibrary/build.gradle
```diff
@@ -3,19 +3,31 @@ apply plugin: "kotlin-android"
...
-  testImplementation deps.test.junit
-  testImplementation deps.test.kotlinTest
+    api deps.external.kotlinStdlib
+    implementation 'com.android.support:appcompat-v7:27.1.1'
+    implementation 'com.android.support.constraint:constraint-layout:1.0.2'
+    implementation 'com.android.support:design:27.1.1'
+    testImplementation deps.test.junit
+    testImplementation deps.test.kotlinTest
+
+    androidTestImplementation deps.test.junit
```
kageiit: This is a breaking change/behavior for okbuck consumers

We should confirm that `androidTest` configurations do not extend from `test` configurations in AGP

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidBuckRuleComposer.groovy
```diff
@@ -34,7 +34,7 @@ abstract class AndroidBuckRuleComposer extends JvmBuckRuleComposer {
...
+    static String bin(AndroidTarget target) {
```
kageiit: Having `bin_` for android libraries does not make sense. Lets name these differently since the `bin_` is supposed to be for actual apps.

Lets call these rules with name `lib_instrumentation_${target.name}`

libraries/kotlinandroidlibrary/build.gradle
thalescm: They shouldn't. There are example of libraries you can use on jvm tests that you can't on android runtime, as ART doesn't have some java packages (such as java.nio.file) But I'll confirm

thalescm: But why is this a breaking chage/behavior? Do we assume they do?

kageiit: Before some dependencies were picked up from `test` configs, but now they wont be. This might require some consumers to add such dependencies in the `androidTest` configurations explicitly. So its a potential breaking change

thalescm: Hum, I've confirmed and test and androidTest configs do not extend from each other. So the right thing to do (and consumers should be doing that already, if they wanna make their project "gradle supported" too), is to explicitly put those configs. WDYT?

kageiit: Ya I agree, its the right thing to do. We just need to call out in the release notes that the behavior changed and its still breaking for someone that relied on the wrong behavior before :)

Just making a note to add it to the release notes for myself. The change itself is fine

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidBuckRuleComposer.groovy
thalescm: This `bin` method is called on [here](https://github.com/uber/okbuck/blob/master/buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidBinaryRuleComposer.groovy#L63).
Though of two solutions, either passing a possibly custom name for this composer, or create a function named bin, which accepts the androidLibInstrumentation target, that will output the `lib_instrumentation` name

thalescm:
@kageiit If you mind taking another look :)

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidBuckRuleComposer.groovy
```diff
@@ -38,6 +40,10 @@ abstract class AndroidBuckRuleComposer extends JvmBuckRuleComposer {
         return "bin_${target.name}"
     }
 
+    static String bin(AndroidLibInstrumentationTarget target) {
```
kageiit: rename method to `libInstrumentation`

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidResourceRuleComposer.groovy
```diff
@@ -13,7 +13,7 @@ final class AndroidResourceRuleComposer extends AndroidBuckRuleComposer {
         // no instance
     }
 
-    static Rule compose(AndroidTarget target) {
+    static Rule compose(AndroidTarget target, List<String> extraDeps = []) {
```
kageiit: rename to `extraResDeps`

buildSrc/src/main/groovy/com/uber/okbuck/core/model/android/AndroidLibInstrumentationTarget.groovy
```diff
@@ -0,0 +1,41 @@
...
+
+import com.android.build.gradle.api.BaseVariant
+import com.android.build.gradle.api.LibraryVariant
+import com.android.builder.model.SourceProvider
+import org.gradle.api.Project
+
+/**
+ * An Android library instrumentation target
+ */
+class AndroidLibInstrumentationTarget extends AndroidAppInstrumentationTarget {
```
kageiit: This is a bit counterintuitive.

The app target extending the lib target makes more sense

kageiit: We should make a base `InstrumentationTarget` and add overrides in app and lib target classes

buildSrc/src/main/rocker/com/uber/okbuck/template/android/RobolectricRule.rocker.raw
```diff
@@ -0,0 +1,50 @@
+@import java.util.Collection
```
kageiit: Lets rename to `AndroidTestRule` since in the future, there will be such rules that dont use robolectric as well

buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidBuckRuleComposer.groovy
thalescm: If this is renamed, on the actual rule it will be named "bin_${target.name}" due to [this](https://github.com/uber/okbuck/blob/master/buildSrc/src/main/groovy/com/uber/okbuck/composer/android/AndroidBinaryRuleComposer.groovy#L63)

buildSrc/src/main/rocker/com/uber/okbuck/template/android/RobolectricRule.rocker.raw
thalescm: I'd call it `AndroidUnitTestRule`, as `AndroidTest` can be cause confusion, as in gradle it means instrumentation tests

kageiit: AGP source code is different from okbuck source code. To keep with the conventions of the other classes, lets use `AndroidTest`

thalescm:
@kageiit If you mind taking yet another look 😬

kageiit:
The dependencies before and after do not look right

screen shot 2018-04-09 at 1 49 28 pm
screen shot 2018-04-09 at 1 49 37 pm

There is a lot of new provided deps. @thalescm can you check please? This looks like a bug

thalescm:
It does indeed. I'll take a look

thalescm:
Checked here, by having
```
implementation 'com.android.support:appcompat-v7:27.1.0'
implementation 'com.android.support.constraint:constraint-layout:1.0.2'
implementation 'com.android.support:design:27.1.0'
```
on `kotlinandroidlibrary/build.gradle` (which was already wrong, it was auto generated when I created the new activity and I didn't pay attention to it, my bad), as the support version was different from the `app` one, it appeared as provided deps (I honestly don't know why).
Changing to the project convention, using everything declared on `dependencies.gradle` this seems to be fixed.

kageiit:
CI fails
```
2018-04-09 21:20:54,093 [WARNING][buck_tool.py:404] Not using buckd because watchman isn't installed.
PARSING BUCK FILES... 0.1s (100%)
BUILD FAILED: The rule //.okbuck/gen/libraries/kotlinandroidlibrary/debug_test:app.keystore could not be found.
Please check the spelling and whether it exists in /home/circleci/okbuck/.okbuck/gen/libraries/kotlinandroidlibrary/debug_test/BUCK.
```

thalescm:
Strange, this works on my machine even after full cleans:
```
$ rm -rf ./buck-out/ && rm -rf ./.okbuck/ && ./gradlew clean && ./buckw build //libraries/kotlinandroidlibrary:lib_instrumentation_debug_test_apk

Deprecated Gradle features were used in this build, making it incompatible with Gradle 5.0.
See https://docs.gradle.org/4.6/userguide/command_line_interface.html#sec:command_line_warnings

BUILD SUCCESSFUL in 1s
17 actionable tasks: 4 executed, 13 up-to-date
NO PREVIOUS SUCCESSFUL OKBUCK RUN
RUNNING OKBUCK...

<Ok buck setup logs>

BUILD SUCCESSFUL in 10s
24 actionable tasks: 19 executed, 5 up-to-date

PROCEEDING WITH BUCK

Action graph will be rebuilt because files have been added or removed.
Parsing buck files: finished in 0.6 sec (100%)
Building: finished in 12.2 sec (100%) 224/224 jobs, 224 updated, 98.2% cache miss
  Total time: 12.8 sec
```

PR #625, opened 2018-04-10
Add classifier to version string when creating external dependency object from an artifact.

raviagarwal7:
Added tests for usage of classifiers.


buildSrc/src/main/java/com/uber/okbuck/core/model/base/Scope.java
```diff
@@ -318,13 +319,25 @@ private void extractConfiguration(Configuration configuration) {
             ComponentIdentifier identifier = artifact.getId().getComponentIdentifier();
             artifactIds.add(identifier);
 
+            String classifier = null;
+            if (artifact.getId() instanceof DefaultModuleComponentArtifactIdentifier) {
+                classifier = ((DefaultModuleComponentArtifactIdentifier) artifact.getId())
```
kageiit: cc @oehme anyway we can get the classifier without relying on internal gradle apis?

oehme: There's no api, since classifiers will eventually disappear from Gradle's model (replaced by variant attributes). You could look at `artifactIdentifier.fileName`, the classifier is the last segment before the file extension. Both approaches should be fine for quite a while, since getting rid of "classifier" will be a longer process.

kageiit: Thanks for clarifying. We were thinking of using the file approach, but we're not sure if a more direct API existed. We will switch to using variant attributes when it's available and use the filene methodology for now. Thanks!

PR #647, opened 2018-05-02
Populate maven_coords attribute for prebuilt jars and aars in dep cache

romanoid:


buildSrc/src/main/java/com/uber/okbuck/core/dependency/ExternalDependency.java
```diff
@@ -57,7 +58,7 @@ public String getCacheName() {
     public String getCacheName(boolean useFullDepName) {
         if (useFullDepName) {
             if (!StringUtils.isEmpty(group)) {
-                return group + "." + depFile.getName();
+                return group + "--" + name + "--" + String.valueOf(version) + "." + packaging;
```
kageiit: lets move `--` to a constant so we can change it easier if needed

romanoid: done

buildSrc/src/main/resources/com/uber/okbuck/core/util/thirdparty/BUCK_FILE
```diff
@@ -21,17 +21,25 @@ for aar in glob(['*.aar']):
 
 for jar in jars:
   basename = base_name_regex.sub('', jar)
+
+  maven_coords = None
+  if ('--' in basename):
```
kageiit: make '--' a constant as well

PR #656, opened 2018-05-16
Add GJF code style pre commit hook support.

raviagarwal7:


artem-zinnatullin:
Have you checked if it works with `-a`? 

ie:

```console
# Do some changes.
$ git commit -am 'commit message'
```

I was doing pretty much same thing to autoformat code, but faced problem with committing with `-a` flag. 

I don't remember the details exactly, but I think `-a` creates secondary temporary `.git/index` and if you try to add something to git index in the hook, it breaks its internal state and both index files remain and git starts erroring

PR #666, opened 2018-05-23
Add Builder for Scope

kageiit:


buildSrc/src/main/java/com/uber/okbuck/core/model/android/AndroidTarget.java
```diff
@@ -157,22 +157,24 @@ public AndroidTarget(Project project, String name) {
...
-        getUnitTestVariant() != null ? getSources(getUnitTestVariant()) : ImmutableSet.of(),
-        getJavaResources(getUnitTestVariant()),
-        getJavaCompilerOptions(getUnitTestVariant()));
+    Scope.Builder builder = Scope.builder(getProject());
+    UnitTestVariant unitTestVariant = getUnitTestVariant();
+    if (unitTestVariant != null) {
+      builder.configuration(unitTestVariant.getRuntimeConfiguration());
+      builder.sourceDirs(getSources(getUnitTestVariant()));
+      builder.javaCompilerOptions(getJavaCompilerOptions(unitTestVariant));
+    }
```
raviagarwal7: missing `resources`

kageiit: good catch. mind fixing in a follow up?

kageiit: Fixed in https://github.com/uber/okbuck/commit/1d25f8918783a3870d46709159f6101fc94fbf8a

kageiit: for null, we return empty map which we dont need to handle anymore as its the default. Will clean up that code later

PR #678, opened 2018-06-06
Improved classifier support for generated mvn_coords.

romanoid:
- Updated classifier support to replace classifier being suffix of version into its own entity.
- Added parsing of classifier into dependency cache BUCK file


raviagarwal7:
LGTM!

please verify that the generated buck files are same before/after this change in android repo.

PR #699, opened 2018-08-21
Enable errorprone and nullaway on okbuck codebase

### LINKED ISSUE #692: Add ErrorProne to okbuck
https://github.com/google/error-prone/blob/master/examples/gradle/build.gradle

kageiit:
Fixes #692


buildSrc/build.gradle
```diff
@@ -56,6 +65,12 @@ rocker {
     extendsModelClass "com.uber.okbuck.template.core.Rule"
 }
 
+tasks.withType(JavaCompile) {
+  if (!name.toLowerCase().contains("test")) {
```
hzsweers: Why not run on tests?

buildSrc/src/main/java/com/uber/okbuck/OkBuckGradlePlugin.java
```diff
@@ -81,7 +81,10 @@
   BuckManager buckManager;
 
   // Only apply to the root project
-  public void apply(@NotNull Project rootProject) {
+  @Initializer
+  @SuppressWarnings("NullAway")
```
hzsweers: Shouldn't you only need one or the other?

buildSrc/src/main/java/com/uber/okbuck/core/manager/GroovyManager.java
```diff
@@ -16,6 +16,8 @@
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.Configuration;
 
+import javax.annotation.Nullable;
```
hzsweers: Might want to standardize one. I see both this and jetbrains' in this PR

buildSrc/src/main/java/com/uber/okbuck/core/model/jvm/JvmTarget.java
```diff
@@ -268,7 +269,7 @@ public static String javaVersion(JavaVersion version) {
                 + File.separator
                 + "kotlin-allopen.jar");
 
-        for (String annotation : allOpenAnnotations.split(",")) {
```
hzsweers: TIL

buildSrc/src/main/java/com/uber/okbuck/core/task/OkBuckCleanTask.java
```diff
@@ -39,11 +40,12 @@ void clean() throws IOException {
...
     if (okbuckState.exists()) {
-      lastProjectPaths =
-          Files.lines(okbuckState.toPath())
-              .map(String::trim)
-              .filter(s -> s.length() > 0)
-              .collect(MoreCollectors.toImmutableSet());
+      try (Stream<String> lines = Files.lines(okbuckState.toPath())) {
+        lastProjectPaths =
+            lines.map(String::trim)
+                .filter(s -> s.length() > 0)
```
hzsweers: `!Strings.isEmpty()`, or even `!Strings.isBlank()` to rule out whitespace-only text

buildSrc/build.gradle
kageiit: There are no tests

I guess it can be removed :)

buildSrc/src/main/java/com/uber/okbuck/OkBuckGradlePlugin.java
kageiit: Will remove initializer

buildSrc/src/main/java/com/uber/okbuck/core/manager/GroovyManager.java
kageiit: catch. Will fix everywhere

buildSrc/src/main/java/com/uber/okbuck/core/task/OkBuckCleanTask.java
kageiit: Sure

buildSrc/src/main/java/com/uber/okbuck/OkBuckGradlePlugin.java
kageiit: Actually seems like I need both. Otherwise nullaway throws several errors instead of just the one
```
[NullAway] initializer method does not guarantee @NonNull fields manifestPath, packageName are initialized along all control-flow paths (remember to check for exceptions or early returns).
```

Which I am able to suppress with a more targeted and single method suppression

PR #708, opened 2018-09-05
Added Visibility configuration support to OkBuck.

romanoid:
visibility is configured by adding `VISIBILITY.bzl` file in the same directory as `build.gradle/BUCK` file

This file needs to have method `module_visibility` defined that returns visibility.

Example:

```
def module_visibility():
  return [
    "//buildSrc/...",
    "//plugins/...",
  ]
```

### DIFF
--- a/buildSrc/src/main/java/com/uber/okbuck/core/util/FileUtil.java
+++ b/buildSrc/src/main/java/com/uber/okbuck/core/util/FileUtil.java
@@ -3,7 +3,13 @@ package com.uber.okbuck.core.util;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Multimap;
 import com.google.common.collect.TreeMultimap;
+import com.uber.okbuck.template.common.LoadStatements;
 import com.uber.okbuck.template.core.Rule;
+import org.apache.commons.io.FileUtils;
+import org.gradle.api.Project;
+import org.slf4j.Logger;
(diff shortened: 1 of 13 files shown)


buildSrc/src/main/java/com/uber/okbuck/core/util/FileUtil.java
```diff
@@ -79,16 +89,22 @@ public FileVisitResult postVisitDirectory(Path dir, IOException exc)
...
       }
 
-      try {
-        buckFile.createNewFile();
+      try (
+          OutputStream fos = new FileOutputStream(buckFile);
+          BufferedOutputStream os = new BufferedOutputStream(fos)){
 
-        OutputStream os = new FileOutputStream(buckFile);
+        writeLoadStatements(loadStatements, os);
```
kageiit: Can we use rocker for passing in a list of loads instead?

romanoid: I think code will be much clunkier for this use case (since we'll have to use all the same java logic, but instead inside rocker) 

I can be convinced to do this for consistency though

kageiit: Let's do it for consistency please

kageiit:
Can you please enable this feature in this pr by adding an example/integration test

PR #713, opened 2018-09-09
Setup circle ci workflow

kageiit:


alexfdz:
@kageiit I'm trying to understand the motivation behind .circleci/config.yml#L28. Why this specific library is a problem in cache?

kageiit:
We are currently on snapshots of butter knife since there has been no stable release for over a year. Ok uxk can only use one source jar for an artifact and the constant snapshot updating breaks okbuck. This change was done to avoid flakiness in ci

PR #724, opened 2018-09-17
Fix generated exopackage android library rule

raviagarwal7:
**Description**:
- Fix generated exopackage android library rule
- Use exo variant's android manifest for exopackage instead of main's.

### DIFF
--- a/buildSrc/src/main/java/com/uber/okbuck/core/model/android/AndroidAppTarget.java
+++ b/buildSrc/src/main/java/com/uber/okbuck/core/model/android/AndroidAppTarget.java
@@ -181,17 +181,16 @@ public class AndroidAppTarget extends AndroidLibTarget {
   @Nullable
   private String getExoManifest() {
     String mainManifest = getMainManifest();
+    List<String> secondaryManifests = getSecondaryManifests();
 
-    if (mainManifest != null) {
-      String exoManifest = mainManifest.replace("main", "exo");
+    if (secondaryManifests != null) {
+      Optional<String> optionalExoManifest =
+          secondaryManifests.stream().filter(manifest -> manifest.contains(getName())).findAny();
 
-      if (getProject().file(exoManifest).exists()) {
-        return exoManifest;
-      } else {
-        return mainManifest;
-      }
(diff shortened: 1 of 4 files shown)


buildSrc/src/main/java/com/uber/okbuck/core/model/android/AndroidAppTarget.java
```diff
@@ -179,6 +178,22 @@ private Keystore extractKeystore() {
     return null;
   }
 
+  @Nullable
+  private String getExoManifest() {
+    String mainManifest = getMainManifest();
+
+    if (mainManifest != null) {
+      String exoManifest = mainManifest.replace("main", "exo");
```
kageiit: What's the purpose of the string replace. This is very specific and only works if folks name their exo variant 'exo'. Can we not hardcore the names here?

buildSrc/src/main/rocker/com/uber/okbuck/template/config/BuckDefs.rocker.raw
```diff
@@ -187,7 +187,7 @@ def okbuck_android_library(
         **kwargs
     )
 
-    if isLintEnabled():
+    if isLintEnabled() and not disable_lint and manifest is not None:
```
kageiit: Jus lt pass in manifest as a paramedic to isLintEnabled. We do not need to check disable_lint as it is templatized into the function body

raviagarwal7:
addressed review comments.

these are the list of targets seen on a application now
```
//app:app_lib_exo
//app:bin_debug
//app:bin_exo
//app:bin_release
//app:build_config_debug
//app:build_config_exo
//app:build_config_release
//app:keystore_debug
//app:keystore_debug_properties
//app:keystore_exo
//app:keystore_exo_properties
//app:keystore_release
//app:keystore_release_properties
//app:lint_debug
//app:lint_exo
//app:lint_release
//app:manifest_debug
//app:manifest_exo
//app:manifest_release
//app:res_debug
//app:res_exo
//app:res_release
//app:src_debug
//app:src_exo
//app:src_release
```

PR #728, opened 2018-09-26
Add secondary manifest to okbuck_manifest rule

raviagarwal7:
This fixes manifest genrule to generate correct merged manifest for flavors.


buildSrc/src/main/java/com/uber/okbuck/composer/jvm/JvmLibraryRuleComposer.java
```diff
@@ -10,6 +10,7 @@
 import java.util.List;
 import java.util.Set;
 
+@SuppressWarnings("NullAway")
```
kageiit: Why is this related?

raviagarwal7: not sure, nullaway threw errors on
```
          new JvmBinaryRule()
              .mainClassName(target.getMainClass())
```
target.getMainClass() could be null

not sure why it isn't wasn't caught before

PR #731, opened 2018-09-29
Add ability to configure buck's resource_union package name

kageiit:
**Description**:
- Change `resourceUnion` extension option to `resourceUnionPackage`
- Use resource union package when available for library manifests (IDE support etc.)
- Make sure binaries still use the final application id
- Cleanup some okbuck defs

### DIFF
--- a/buildSrc/src/main/rocker/com/uber/okbuck/template/android/ManifestRule.rocker.raw
+++ b/buildSrc/src/main/rocker/com/uber/okbuck/template/android/ManifestRule.rocker.raw
@@ -16,8 +16,6 @@ okbuck_manifest(
     target_sdk = '@targetSdk',
 @if (valid(pkg)) {
     package = '@pkg',
-} else {
-    package = '',
 }
 @if (valid(versionCode)) {
     version_code = '@versionCode',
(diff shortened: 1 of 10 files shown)


buildSrc/src/main/rocker/com/uber/okbuck/template/android/ManifestRule.rocker.raw
```diff
@@ -14,14 +14,16 @@ okbuck_manifest(
     main_manifest = '@mainManifest',
     min_sdk = '@minSdk',
     target_sdk = '@targetSdk',
+@if (valid(pkg)) {
+    package = '@pkg',
+} else {
+    package = '',
```
raviagarwal7: probably can skip passing package. does it replace the package with an empty string?

kageiit: Done

alexfdz:
When are facing several issues upgrading from `okbuck:0.41.0` to `okbuck:0.42.0`, and probably are related to this issue.

We are in a scenario of an app with several 3rd party android libraries, and multiple combinations of build types and product flavors.

With the upgrade, we get lots of compilation errors in the android libraries of the style:

```
error: cannot find symbol
symbol:  _variable_name
location: class com.external.library.libraryA.R.string
```
`_variable_name` is a string resource defined in a different module (`libraryB`) and is a dependency of `libraryA`.

The problem is that we want to keep the `resource_union` behaviour, but we don't want to declare one specific `resourceunionPackage` as each module has its own.

Did I get the issue correctly? Can you think about any workaround?

PR #732, opened 2018-10-02
Refactor package code & fix disable lint logic

raviagarwal7:
- Build Config
-- Use application id with suffix for APPLICATION_ID
-- Use base application id for libraries (since this can't be overridden via dsl) & original package for apps. 
- Manifest
-- Use res package (i.e use if defined else use original package from manifest) for libraries
-- Use application id with suffix for apps
- Don’t generate lint rules for excluded buildTypes


buildSrc/src/main/java/com/uber/okbuck/composer/android/AndroidLibraryRuleComposer.java
```diff
@@ -83,6 +83,7 @@ public static Rule compose(
       androidRule.language("kotlin");
     }
 
+    androidRule.isLintTarget(true);
```
kageiit: whats the use of this flag?

raviagarwal7: this is to skip generating any lint arguments for android test targets since it depends on android rule target.

PR #734, opened 2018-10-02
Add support for API 28 robolectric dependencies expected by robolectric 4.0-alpha-3

alexfdz:
**Description**: Robolectric 4.0-alpha-3 is their first release with support for API 28, and it requires the SDK dependency `android-all-9-robolectric-4799589` as specified in their [org.robolectric.internal.SdConfig](https://github.com/robolectric/robolectric/blob/4fbb097ff619fab3ea405e1a1ce7b04a83dfe836/robolectric/src/main/java/org/robolectric/internal/SdkConfig.java#L31).

Currently okbuck specifies as sdk dependency for API 28 `android-all-P-robolectric-4651975`.

This means that you cannot use okbuck with an API 28 with robolectric, as robolectric expects `android-all-9-robolectric-4799589` but okbuck only offers `android-all-P-robolectric-4651975`.

### DIFF
--- a/buildSrc/src/main/java/com/uber/okbuck/core/manager/RobolectricManager.java
+++ b/buildSrc/src/main/java/com/uber/okbuck/core/manager/RobolectricManager.java
@@ -82,7 +82,7 @@ public final class RobolectricManager {
     API_25("7.1.0_r7", "r1"),
     API_26("8.0.0_r4", "r1"),
     API_27("8.1.0", "4611349"),
-    API_28("P", "4651975");
+    API_28("9", "4799589");
 
     private final String androidVersion;
     private final String frameworkSdkBuildVersion;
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/java/com/uber/okbuck/core/manager/RobolectricManager.java
```diff
@@ -82,7 +82,7 @@ public void finalizeDependencies() {
     API_25("7.1.0_r7", "r1"),
     API_26("8.0.0_r4", "r1"),
     API_27("8.1.0", "4611349"),
-    API_28("P", "4651975");
+    API_28("9", "4799589");
```
kageiit: We cannot do this for backwards compatibility reasons. Lets create a new enum API_28 and move the existing one to API_P so we can get both the jars

alexfdz: Good point @kageiit, just fixed it.

PR #742, opened 2018-10-15
Remove compilation classpath from lint targets by default

kageiit:
While lint does have the ability to do bytecode analysis (which would require compiling up front), nearly all the bytecode-based checks have been rewritten as source-based checks instead (which has other nice attributes, such as more accurate source position information and being able to run on the fly in the editor for not-yet- or un- compileable code)

This changes okbuck generated lint targets to not have to depend on compiling the target before running lint and thus speeds up overall build times.

There is an option to retain the old behavior incase lint needs to run on generated code which is generated during annotation processing.


buildSrc/src/main/rocker/com/uber/okbuck/template/config/BuckDefs.rocker.raw
```diff
@@ -104,7 +105,10 @@ def okbuck_lint(
     bash += "-classpath {} com.android.tools.lint.Main ".format(toLocation('//.okbuck/workspace/lint:okbuck_lint'))
 
     if has_srcs:
-        bash += "--classpath $(location :src_{}) --libraries `cat $CP_FILE` ".format(variant)
+@if(useCompilationClasspath) {
+        bash += "--classpath $(location :src_{}) "
```
raviagarwal7: `.format(variant)` should be here

raviagarwal7: should we also add `bash += "CP_FILE=`sed 's/@@//' <<< $(@@@(classpathMacro) :src_{})`; ".format(variant)` conditionally using `useCompilationClasspath`, since classpath abi will require compilation as well

kageiit: classpath abi does not include abi of src_lib target does it?

PR #745, opened 2018-10-20
Fix project-defined annotation processor resolution

artem-zinnatullin:
Problem:

If project defines custom annotation processor, it needs to be declared in `okbuck.annotationProcessors`, this in turn makes OkBuck `!ProjectUtil.getAnnotationProcessorCache(getProject())
        .hasEmptyAnnotationProcessors(getProject(), configurationName)) {
        .hasEmptyAnnotationProcessors(getProject(), configurationName)` return true and results in unconfigured scope being passed to rule renderer.

TL;TR:

without fix we're getting `BUCK` file with following content

```python
okbuck_android_library(
    name = 'src_release',
    srcs = glob([
        'src/main/java/**/*.java',
        'src/main/java/**/*.kt',
    ]),
    source = '8',
    target = '8',
    tests = [
        ':test_release',
    ],
    plugins = [
        '//.okbuck/workspace/processor:processor_com.lyft.dagger1-dagger-compiler-1.2.6',
        '//.okbuck/workspace/processor:processor_instant-android.instant-processors.instant-bind-dagger1-unspecified',
    ],
)
```

Note that `annotation_processor_deps` is missing here.

With fix we're getting correct `BUCK` file:

```python
okbuck_android_library(
    name = 'src_release',
    srcs = glob([
        'src/main/java/**/*.java',
        'src/main/java/**/*.kt',
    ]),
    source = '8',
    target = '8',
    tests = [
        ':test_release',
    ],
    plugins = [
        '//.okbuck/workspace/processor:processor_com.lyft.dagger1-dagger-compiler-1.2.6',
        '//.okbuck/workspace/processor:processor_instant-android.instant-processors.instant-bind-dagger1-unspecified',
    ],
    annotation_processor_deps = [
        '//.okbuck/ext/com/google/auto:auto-common-0.6.jar',
        '//.okbuck/ext/com/google/guava:guava-20.0.jar',
        '//.okbuck/ext/com/lyft/dagger1:dagger-1.2.6.jar',
        '//.okbuck/ext/com/lyft/dagger1:dagger-compiler-1.2.6.jar',
        '//.okbuck/ext/com/squareup:javapoet-1.9.0.jar',
        '//.okbuck/ext/javax/inject:javax.inject-1.jar',
        '//instant-frameworks/logging:src_main',
        '//instant-processors/instant-bind/annotations:src_main',
        '//instant-processors/instant-bind/dagger1:src_main',
        '//instant-processors/instant-bind/processor-common:src_main',
    ],
)
```


kageiit:
cc @raviagarwal7

raviagarwal7:
let me take a look. We don't want to `annotation_processor_deps` if there are no empty annotation processors. i.e all the dependencies are coming from plugin and need not be defined in `annotation_processor_deps`

artem-zinnatullin:
I think your intention is right, but I'm not sure the code is doing exactly that

It seems that if you have empty annotation processors (moreover, it probably should only be the case if all of them are empty), then you want to configure the project with no actual deps configuration which is what this PR tries to achieve

It seems like the condition in the code was mistakenly inverted or I'm missing something :)

raviagarwal7:
> I think your intention is right, but I'm not sure the code is doing exactly that
> 
> It seems that if you have empty annotation processors (moreover, it probably should only be the case if all of them are empty), then you want to configure the project with no actual deps configuration which is what this PR tries to achieve
> 
> It seems like the condition in the code was mistakenly inverted or I'm missing something :)

It's the opposite. 

If there are any empty annotation processors in the `annotationProcessor` configuration that means they won't get included in the plugin's dependency list. hence the dependencies needs to be put in the `annotation_processor_deps` list so that it is available at annotation processor's classpath when the ap runs.

but when there are no empty annotation processors, there is no need to generate the `annotation_processor_deps` list.

eg: is https://github.com/uber/okbuck/blob/master/app/build.gradle#L75. removing `annotationProcessor deps.apt.autoValueAnnotations` leads to no `annotation_processor_deps` list
this is because each of the generated plugin's dependencies are sufficient for running the annotation processor. 
```
        '//.okbuck/workspace/processor:processor_com.google.auto.value-auto-value-1.6__525046afa7fcbcf1dcd36f17908d83ff',
        '//.okbuck/workspace/processor:processor_com.google.dagger-dagger-compiler-2.15',
        '//.okbuck/workspace/processor:processor_com.jakewharton-butterknife-compiler-9.0.0-SNAPSHOT',
    ],
```
-------

do you have a use case you can share which breaks due to the current code?

artem-zinnatullin:
hm hm

>do you have a use case you can share which breaks due to the current code?

Yeah, in our case we use 3rd-party and own annotation processors, our own annotation processors are part of the project Buck is building. 

Unfortunately, Buck doesn't recognize our aps automatically so we have to declare them in `okbuck { annotationProcessors = … }`. We had this code for long time.

Updating OkBuck from `0.37.4` to `0.43.0` results in build failures, particularly because we're not getting aps properly configured (examples of `BUCK` files are in the comment above).

kageiit:
It would help if there was a small sample project showing the issue

artem-zinnatullin:
Working on it

artem-zinnatullin:
Damn it, can't reproduce… Pushed integration projects that I've used as tests

artem-zinnatullin:
Gonna dive deeper into cause of the build failure in our project

artem-zinnatullin:
Ohhh, looks like it happens in modules that use Kotlin

If I manually set `annotation_processing_tool = 'javac'` in that module's `BUCK` rules, it builds fine.

There was a detail I forgot to mention, the failing `javac` invocation has `-proc:none` which if I'm not mistaken Buck would set in two cases:

- No APs were explicitly defined
- Processing is delegated to KAPT (which is default for Kotlin Modules)

Gonna try to reproduce with sample projects in OkBuck

artem-zinnatullin:
Ok, something is really off if module uses Kotlin.

I've pushed another sample project that has Java code that needs AP, but module also configured as Kotlin

Generated `BUCK` doesn't mention ap at all now:

```python
lib_deps = []

kotlin_library(
    name = 'src_main',
    srcs = glob([
        'src/main/java/**/*.java',
        'src/main/java/**/*.kt',
    ]),
    source = '8',
    target = '8',
    provided_deps = [
        '//.okbuck/ext/com/google/auto/value:auto-value-annotations.jar',
        '//libraries/annotationprocessor:src_main',
    ],
    maven_coords = 'okbuck.libraries:annotationprocessorconsumerkotlin:unspecified',
    visibility = [
        'PUBLIC',
    ],
)
```

That's not what happens in my case though, I'm pretty confused now.

raviagarwal7:
thanks @artem-zinnatullin for adding the kotlin module. let me take a look.

artem-zinnatullin:
We got the latter problem reproduced in our project today, applying Kotlin Gradle plugin to java-library module results in `BUCK` file rendered without ap

Before:

```python
lib_deps = []

java_library(
    name = 'src_main',
    srcs = glob([
        'src/main/java/**/*.java',
    ]),
    source = '8',
    target = '8',
    tests = [
        ':test_main',
    ],
    plugins = [
        '//.okbuck/workspace/processor:processor_com.lyft.dagger1-dagger-compiler-1.2.7',
    ],
    maven_coords = 'instant-android.instant-features.passenger:ride:unspecified',
    deps = [
        '//.okbuck/ext/com/android/support:support-annotations-28.0.0.jar',
        '//.okbuck/ext/com/google/code/findbugs:jsr305-3.0.2.jar',
        '//.okbuck/ext/com/google/code/gson:gson-2.8.5.jar',
        // … goes on
    ],
    visibility = [
        'PUBLIC',
    ],
)
```

After applying `kotlin` plugin:

```python
kotlin_library(
    name = 'src_main',
    srcs = glob([
        'src/main/java/**/*.java',
        'src/main/java/**/*.kt',
    ]),
    source = '8',
    target = '8',
    tests = [
        ':test_main',
    ],
    maven_coords = 'instant-android.instant-features.passenger:ride:unspecified',
    deps = [
        '//.okbuck/ext/com/android/support:support-annotations-28.0.0.jar',
        '//.okbuck/ext/com/google/code/findbugs:jsr305-3.0.2.jar',
        '//.okbuck/ext/com/google/code/gson:gson-2.8.5.jar',
        // … goes on
    ],
    visibility = [
        'PUBLIC',
    ],
)

```

libraries/annotationprocessorconsumerkotlin/build.gradle
```diff
@@ -0,0 +1,10 @@
+apply plugin: "kotlin"
+
```
raviagarwal7:
```suggestion
apply plugin: "kotlin-kapt"
```

you need to use `kapt` for kotlin libraries. annotationProcessor doesn't work.

libraries/annotationprocessorconsumerkotlin/build.gradle
```diff
@@ -0,0 +1,10 @@
...
+dependencies {
+    annotationProcessor project(":libraries:annotationprocessor")
+    annotationProcessor deps.apt.autoValue
+    annotationProcessor project(":libraries:annotationprocessoroverexistingap")
```
raviagarwal7: replace `annotationProcessor` usages with `kapt` and I am able to see the plugin argument generated. the build was successful as well.

raviagarwal7:
the issue is with mixed sources. the kotlin example that you added has java sources which fails to build.

this is how we create a jvm target (https://github.com/uber/okbuck/blob/master/buildSrc/src/main/java/com/uber/okbuck/core/model/base/TargetCache.java#L52), it could either be a kotlin target or plain java target. 

for kotlin target we just take `kapt` configuration and `annotationProcessor` configuration is ignored. which is why the `plugin` argument was not getting generated in the buck file.

If we want this to work we would need to make `kotlin_library` in buck pass in annotation processors plugin when building the java sources.

raviagarwal7:
> After applying `kotlin` plugin:
Use `kapt` if applying `kotlin` plugin.

artem-zinnatullin:
We deliberately don't use kapt for performance and stability reasons see https://github.com/facebook/buck/pull/1987 and only run javac as apt

Could OkBuck fallback to this behavior if kapt Gradle plugin is not applied to the module?

kageiit:
> We deliberately don't use kapt for performance and stability reasons see [facebook/buck#1987](https://github.com/facebook/buck/pull/1987) and only run javac as apt
> 
> Could OkBuck fallback to this behavior if kapt Gradle plugin is not applied to the module?

Unfortunately, that is not possible since kapt and annotatiopnProcessor are two completely different configurations. The usecase you describe seems to be a workaround that would compromise correctness in okbuck and cannot be supported. I would recommend you do instead use kapt by default, but when `-Dokbuck.wrappper` is not set, change all the deps to `annotationProcessor` instead via some gradle configuration instead and null out the kapt config.

artem-zinnatullin:
I think I didn't explain it properly, 

we don't have deps declared as `kapt` configuration, kapt plugin is not applied. it's pure java apt and `annotationProcessor` configuration all the time

kageiit:
Ah I see what you are saying. This does need a change in okbuck. The flow would be as follows
- Java projects nothing changes
- Kotlin projects
```
- if kapt plugin is applied, okbuck only considers kapt configuration (kotlin shouts at you with a warning anyways). We also set the apt tool to kapt explicitly (even if its the implicit buck default) in the generated buck rule
- if kapt plugin is not applied, we use annotaionProcessor config, but we set the apt tool to javac in the generated buck rule.
```

This should help fix the problem. @artem-zinnatullin would welcome a PR if you want to tackle this :)

artem-zinnatullin:
Glad we have understanding now :)

I'll see how it goes with prioritization, I may eventually implement it but it's free for others to work on, just let us know in the comments :)

PR #747, opened 2018-10-22
Fix lint broken on resources

kageiit:
This fixes android lint by making sure resources are passed in correctly.

Also added integration tests for resource lints, since we only have them for classes till now


buildSrc/src/main/rocker/com/uber/okbuck/template/config/BuckDefs.rocker.raw
```diff
@@ -115,7 +115,7 @@ def okbuck_lint(
     for src in srcs:
         bash += '--sources {}/{} '.format(get_base_path(), src)
 
-    for resource in resources:
+    for resource in res:
```
raviagarwal7: nit: `res_dirs` instead of `res` and `res_dir` instead of `resource`

kageiit: Will fix in follow up

kageiit: Fixed in https://github.com/uber/okbuck/commit/c76e346e29445ce8f3b49aca1ccb6ef1eb41b9e7

raviagarwal7: thanks 👍

PR #749, opened 2018-10-31
Jetifier support

thalescm:
Jetifier support for ok buck:

It will work like this:
* Check if androidX and jetifier is enable (gradle.properties)
* If it is not, do nothing new
* If enabled, download jetifier and it's dependencies
* Created a new def called `okbuck_android_prebuilt_aar` that equals to `android_prebuilt_aar` if jetifier disabled, and to jetify `aars` if enabled.

-------------------------
Next steps

Incoming later is the support to identify which aars are really suitable to be jetified, by checking it's pom dependencies.


buildSrc/src/main/java/com/uber/okbuck/core/manager/JetifierManager.java
```diff
@@ -0,0 +1,107 @@
...
+    private static final ImmutableList<String> JETIFIER_MODULES =
+            ImmutableList.of(
+                    "jetifier-core",
+                    "jetifier-processor"
+            );
+    private static final ImmutableList<String> INTERNAL_MODULES =
+            ImmutableList.of(
+                    "commons-cli.jar",
+                    "jetifier-standalone.jar"
+            );
```
raviagarwal7: these jars doesn't come coupled with jetifier?

tooling/buck-defs/DEFS
```diff
@@ -7,3 +7,25 @@ def android_binary(
...
   )
+
+
+original_android_prebuilt_aar = android_prebuilt_aar
+def android_prebuilt_aar(
+    name,
+    aar,
+    **kwargs
+    ):
+  jetified = name.replace('.jar', '') + "-jetified.aar"
```
raviagarwal7: probably move to a variable. also should be `.aar` instead of `.jar`

buildSrc/src/main/java/com/uber/okbuck/core/manager/JetifierManager.java
```diff
@@ -0,0 +1,107 @@
...
+    private static final String JETIFIER_LOCATION =
+            OkBuckGradlePlugin.WORKSPACE_PATH + "/jetifier";
+    private static final String JETIFIER_BUCK_FILE = JETIFIER_LOCATION + "/BUCK";
+    private static final String JETIFIER_DEPS_CONFIG = "okbuck_jetifier_deps";
+    private static final String JETIFIER_CACHE_LOCATION = "com/android/tools/build/jetifier";
+    private static final String JETIFIER_GROUP = JETIFIER_CACHE_LOCATION
+            .replace("/", ".");
+    private static final String JETIFIER_CLI_CLASS = "com.android.tools.build.jetifier.standalone.Main";
+    private static final String JETIFIER_BINARY_RULE_NAME = "okbuck_jetifier";
+    private static final String JETIFIER_VERSION = "1.0.0-beta02";
```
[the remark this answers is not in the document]
thalescm: @raviagarwal7 I'm marking you here as this also answers your question about the jars.
My first attempt was to make the version configurable, and with the default one being the one that is in the gradle classpath.
The problem is, on google maven, there is only the `jetifier-processor` and `jetifier-core` jars. To use the the default way, that is, the API described [here](https://developer.android.com/studio/command-line/jetifier) it needs those two more jars (if you check the content of the stand alone tool download they're there).
I've made a little experiment to use the gradle defined version (or user defined), which in this case was `1.0.0-alpha02` with those jars pointing to the verion `1.0.0-beta02` and they were incompatible. I've also tried to change the version on the provided link to download the standalone to use `alpha02` but the link was not found, which means that there are huge chances of user defining versions and this not working.
I've then decided it's best to keep the version well defined within the project (which isn't the desirable solution), and at least we can assure everything works and don't start receiving issues do to possible incoherent configurations.
If you have better ideas on how to handle it, please do share :D

buildSrc/src/main/java/com/uber/okbuck/core/manager/JetifierManager.java
```diff
@@ -0,0 +1,115 @@
...
+                    JetifierExtension.JETIFIER_VERSION);
+        }
+
+        Configuration jetifierConfig = project.getConfigurations().maybeCreate(JETIFIER_DEPS_CONFIG);
+        DependencyHandler handler = project.getDependencies();
+        JETIFIER_MODULES
+                .stream()
+                .map(module -> String.format("%s:%s:%s", JETIFIER_GROUP, module, version))
+                .forEach(dependency -> handler.add(JETIFIER_DEPS_CONFIG, dependency));
+        handler.add(JETIFIER_DEPS_CONFIG, "commons-cli:commons-cli:1.3.1");
```
kageiit: lets move commons cli to a constant

buildSrc/src/main/java/com/uber/okbuck/core/model/base/RuleType.java
```diff
@@ -42,17 +42,26 @@
...
+        break;
+      case ANDROID_PREBUILT_AAR:
+        insertOkBuckPrefix = true;
         break;
       default:
         break;
     }
-    return buckType.name().toLowerCase();
+
+    String prefix = insertOkBuckPrefix ? "okbuck_" : "";
```
kageiit: move okbuck_ to a constant

buildSrc/src/main/java/com/uber/okbuck/extension/JetifierExtension.java
```diff
@@ -0,0 +1,19 @@
...
+
+import com.uber.okbuck.core.manager.LintManager;
+
+import org.gradle.api.Project;
+
+import javax.annotation.Nullable;
+
+public class JetifierExtension {
+
+    public static final String JETIFIER_VERSION = "1.0.0-beta02";
```
kageiit: rename to DEFAULT_JETIFIER_VERSION

buildSrc/src/main/rocker/com/uber/okbuck/template/config/BuckDefs.rocker.raw
```diff
@@ -252,3 +253,31 @@ def okbuck_keystore(
...
+        aar = ":{}".format(genname),
+        **kwargs
+    )
+} else {
+    android_prebuilt_aar(
+        name = name,
+        aar = aar,
+        **kwargs
+    )
+}
```
kageiit: newline

buildSrc/src/main/rocker/com/uber/okbuck/template/config/BuckDefs.rocker.raw
```diff
@@ -252,3 +253,31 @@ def okbuck_keystore(
...
+        out = jetified,
+        cmd = "$(exe //.okbuck/workspace/jetifier:okbuck_jetifier) -i $SRCS -o $OUT"
+    )
+
+    android_prebuilt_aar(
+        name = name,
+        aar = ":{}".format(genname),
+        **kwargs
+    )
+} else {
```
kageiit: DRY using two ifs on jetifier

dependencies.gradle
```diff
@@ -2,11 +2,14 @@ def versions = [
         androidPlugin      : "3.2.0",
         androidTools       : "26.2.0",
         butterKnife        : "9.0.0-SNAPSHOT",
-        dagger             : "2.15",
-        kotlin             : "1.2.61",
+        dagger             : "2.16",
```
kageiit: lets use 2.19

gradle.properties
```diff
@@ -0,0 +1,2 @@
+android.useAndroidX=true
+android.enableJetifier=true
```
kageiit: newline

dependencies.gradle
thalescm: We agreed to use 2.16 in person (just detailing here why the change wasn't made)

buildSrc/src/main/java/com/uber/okbuck/core/manager/JetifierManager.java
hzsweers: According to the maintainers, they're not supposed to. Why do we need them?

kageiit: We run jetifier as a cli so we need the standalone jar to call into jetifier-core

Issue #750, opened 2018-11-01
String index out of range


tesla1984:
gradle version: 4.7
okbuck version: 0.44.0

when i run `./buckw targets`, failed on task `:okbuck`

i found the [#39](https://github.com/uber/okbuck/issues/39) ,but my modules are all in the project dir, is there any other reason?
**task execution**
```
:setupOkbuck
:analytics:okbuck
:app:okbuck
:application:okbuck
:appperfmonitor:okbuck
:appperfmonitor-noop:okbuck
:avlib:okbuck
:basiclib:okbuck
:common:okbuck
:commonui:okbuck
:componentlib:okbuck
:componentservice:okbuck
:connect:okbuck
:data:okbuck
:domain:okbuck
:hmsagents:okbuck
:keyboardlayout:okbuck
:logger:okbuck
:mars_daemon:okbuck
:monetization:okbuck
:notbadplayer:okbuck
:router-anno-compiler:okbuck
:router-annotation:okbuck
:tinyhttpd:okbuck
:training:okbuck
:uitestcase:okbuck
:uitestcaseannotation:okbuck
:uitestcaseannotationcomplier:okbuck
:permission:permission-lib:okbuck
:okbuckClean
:okbuck FAILED
```

**the exception**
```
String index out of range: -1
org.gradle.api.tasks.TaskExecutionException: Execution failed for task ':okbuck'.Open stacktrace
Caused by: java.lang.StringIndexOutOfBoundsException: String index out of range: -1Close stacktrace
at com.uber.okbuck.core.util.FileUtil.getRelativePath(FileUtil.java:43)
at com.uber.okbuck.OkBuckGradlePlugin.writeExportedFileRules(OkBuckGradlePlugin.java:300)
at com.uber.okbuck.OkBuckGradlePlugin.lambda$apply$0(OkBuckGradlePlugin.java:175)
at org.gradle.api.internal.AbstractTask$TaskActionWrapper.execute(AbstractTask.java:794)
at org.gradle.api.internal.AbstractTask$TaskActionWrapper.execute(AbstractTask.java:761)
at org.gradle.api.internal.tasks.execution.ExecuteActionsTaskExecuter$1.run(ExecuteActionsTaskExecuter.java:124)
at org.gradle.internal.operations.DefaultBuildOperationExecutor$RunnableBuildOperationWorker.execute(DefaultBuildOperationExecutor.java:317)
[...]

PR #751, opened 2018-11-01
Use path relativize method and create buck files relative to root project.

raviagarwal7:
Also sort created file export rules based on their name.

Resolves https://github.com/uber/okbuck/issues/750


kageiit:
Can you add a root level config file to this change as an integration test?

Issue #758, opened 2018-11-10
resource_union keeping the original package name


alexfdz:
related to #731

When are facing several issues upgrading from `okbuck:0.41.0` to `okbuck:0.42.0`, and probably are related to this issue.

We are in a scenario of an app with several 3rd party android libraries, and multiple combinations of build types and product flavors.

With the upgrade, we get lots of compilation errors in the android libraries of the style:

```
error: cannot find symbol
symbol:  _variable_name
location: class com.external.library.libraryA.R.string
```
`_variable_name` is a string resource defined in a different module (`libraryB`) and is a dependency of `libraryA`.

The problem is that we want to keep the `resource_union` behaviour, but we don't want to declare one specific `resourceunionPackage` as each module has its own.

Did I get the issue correctly? Can you think about any workaround?

alexfdz:
One workaround could be to declare an empty string for `resourceunionPackage` and check with `String#isEmpty` in [AndroidTarget#getPackage](https://github.com/uber/okbuck/blob/3e42015ff8083d14b921467cc9a9f930715dcaa6/buildSrc/src/main/java/com/uber/okbuck/core/model/android/AndroidTarget.java#L363)

kageiit:
Can you provide a simple example of what you mean by

```
The problem is that we want to keep the resource_union behaviour, but we don't want to declare one specific resourceunionPackage as each module has its own.
```

alexfdz:
We are in a scenario of an app and few modules (3rd party android libraries), where the modules have different package names:

- app (com.myapp)
  |- module-core (com.3rdparty.module.core)
  |- module-clientA (com.3rdparty.module.clientA)
  |- module-clientB (com.3rdparty.module.clientB)

Because there are dependencies between the modules (module-clientA -> module-core and module-clientB -> module-core) and we generate `BUCK` files for everything, we need the `android_resource` buck rules for these modules to enable `resource_union`.

With `okbuck:0.41.0` everything was fine, but with `okbuck:0.42.0` if we want to enable the resource union behaviour, we need to define a package, that is applied to every `android_resource#package` buck definition in every app and module in our workspace.

The expected `android_resource#package` for modules, is the module package, not a common one.

alexfdz:
@kageiit we would love some feedback about this problem and collaborate to find a solution

kageiit:
Commenting on the PR

kageiit:
Fixed by #759

PR #759, opened 2018-11-10
Enable resource_union but keeping the original package name

alexfdz:
**Description**: Ability to enable the resource_union configuration in buck, but keeping the original package name.

**Related issue(s)**: #758


kageiit:
The PR makes sense, but it is not obvious that it needs to be set to empty to use the original package name. Maybe it would be better to have two options

- `resourceUnion` (boolean) -> If set to true, use resource union with default package
- `resourceUnionPackage` (string) -> If set, resource union is always enabled, but with the given package name

Can you make this change? I can then merge the PR

alexfdz:
@kageiit done! What about docs?

kageiit:
> @kageiit done! What about docs?

Javadocs are fine. Ill add info to the release notes about the new options

PR #761, opened 2018-11-12
Rewrite okbuck DEFS into SKYLARK-compatible way.

romanoid:
Note: this change changes order of keys returned by subdir_glob and res_glob functions.

**Description**:

**Related issue(s)**:


romanoid:
@kageiit please review this change, but do not merge it yet, I'm going to perform rigorous testing of the change first.

romanoid:
After testing this I've noticed no difference in resources without skylark enabled. Will have to test it alongside enabling skylark.

PR #765, opened 2018-11-19
Enable jetifier for jars

thalescm:
Some jars also depends on old android support packages. So running jetifier on jars is also an important step to fully provide support for jetifier and androidX.

In order to provide that, while still being able to have performatic builds, the need to run jetifier is now calculated per dependency, and not settled entirely for a given packaging type (as it was with aars before)

Main changes, then, are:
* Enable jetifier is now calculated per dependency
* Make generated defs clearer for when jetifier is not enabled in project
* Jetifier by default will be enabled only for all types of external dependencies (jars and aars) packagings. But one can optOut (from jars) by configuring `aarOnly` on the jetifier extension
* It's now also possible, for both jars and aars, to provide a list of regexes of dependencies to exclude from being jetified.

### DIFF
--- a/buildSrc/src/main/java/com/uber/okbuck/extension/JetifierExtension.java
+++ b/buildSrc/src/main/java/com/uber/okbuck/extension/JetifierExtension.java
@@ -1,13 +1,72 @@
 package com.uber.okbuck.extension;
 
+import com.google.common.collect.ImmutableSet;
+import com.uber.okbuck.core.dependency.ExternalDependency;
+import com.uber.okbuck.core.manager.JetifierManager;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+import org.gradle.api.Project;
+
 public class JetifierExtension {
 
   public static final String DEFAULT_JETIFIER_VERSION = "1.0.0-beta02";
+  private static final List<String> BLACKLISTED_DEPS =
+      Arrays.asList(
+          "com.android.tools.build.jetifier:jetifier-core",
+          "com.android.tools.build.jetifier:jetifier-processor",
+          "com.google.code.gson:gson",
+          "commons-cli:commons-cli",
+          "org.jdom:jdom2",
+          "org.jetbrains:annotations",
+          "org.ow2.asm:asm-commons",
+          "org.ow2.asm:asm-tree",
+          "org.ow2.asm:asm-util",
+          "org.ow2.asm:asm",
+          "org.jetbrains.kotlin:.*");
 
   /** Jetifier jar version */
   public String version;
 
-  JetifierExtension() {
+  /** Enable jetify to act on aars only */
+  private boolean aarOnly;
+
+  /** Stores the dependencies which are excluded from being jetified. */
+  private List<String> exclude = new ArrayList<>();
+
+  private final boolean enableJetifier;
+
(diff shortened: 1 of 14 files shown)


buildSrc/src/main/java/com/uber/okbuck/extension/JetifierExtension.java
```diff
@@ -1,13 +1,72 @@
...
+import java.util.List;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+import javax.annotation.Nullable;
+import org.gradle.api.Project;
+
 public class JetifierExtension {
 
   public static final String DEFAULT_JETIFIER_VERSION = "1.0.0-beta02";
+  private static final List<String> BLACKLISTED_DEPS =
```
edsilfer: Would make sense to provide this as an external piece of configuration? Otherwise, everytime that we figure out that a dep need to be blacklisted okbuck version will have to be bumped right?

buildSrc/src/main/java/com/uber/okbuck/extension/JetifierExtension.java
```diff
@@ -1,13 +1,72 @@
...
+              .addAll(BLACKLISTED_DEPS)
+              .build()
+              .stream()
+              .map(Pattern::compile)
+              .collect(Collectors.toList());
+    }
+    return excludePatterns;
+  }
+
+  public boolean shouldJetify(String group, String name, String packaging) {
```
edsilfer: Missing JavaDocs?

buildSrc/src/main/java/com/uber/okbuck/extension/JetifierExtension.java
kageiit: lets rename to JETIFIER_DEPS and addd javadoc to mirror what your comment said

buildSrc/src/main/java/com/uber/okbuck/extension/JetifierExtension.java
```diff
@@ -1,13 +1,80 @@
...
+          "org.ow2.asm:asm-util",
+          "org.ow2.asm:asm",
+          "org.jetbrains.kotlin:.*");
 
   /** Jetifier jar version */
   public String version;
 
-  JetifierExtension() {
+  /** Enable jetify to act on aars only */
+  private boolean aarOnly;
```
kageiit: make public

buildSrc/src/main/java/com/uber/okbuck/extension/JetifierExtension.java
```diff
@@ -1,13 +1,80 @@
...
+
+  /** Stores the dependencies which are excluded from being jetified. */
+  private List<String> exclude = new ArrayList<>();
```
kageiit: make public

buildSrc/src/main/java/com/uber/okbuck/extension/JetifierExtension.java
```diff
@@ -1,13 +1,80 @@
...
+
+  /**
+   * Check if this dependency, described by the params, should be jetified, that is, run
+   * jetifier on it before prebuilding it.
+   * @param group - Dependency group
+   * @param name - Dependency name
+   * @param packaging - Packaging type (aar\jar)
+   * @return true if shouldJetify, false otherwise
+   */
+  public boolean shouldJetify(String group, String name, String packaging) {
```
kageiit: might want to check `enableJetifier` before anything else

thalescm:
@kageiit done

PR #766, opened 2018-11-28
Allowing okBuck to pick oldest source jar among many if multiple source jars are present.

romanoid:
Some of the open-source libraries are published with different source jars

This is generally due to people updating pom with new repository to publish to respective different repository

It is easy for gradle cache to conatin both version of the source jars if multiple remote repos are queried by gradle at the same machine.

known examples:
commons-httpclient
j2objc-annotations

**Description**:

**Related issue(s)**:

### DIFF
--- a/buildSrc/src/main/java/com/uber/okbuck/core/dependency/DependencyUtils.java
+++ b/buildSrc/src/main/java/com/uber/okbuck/core/dependency/DependencyUtils.java
@@ -13,9 +13,12 @@ import java.nio.file.FileSystem;
 import java.nio.file.FileSystems;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.util.Comparator;
 import java.util.Objects;
 import java.util.Set;
+import java.util.jar.JarFile;
 import java.util.stream.Collectors;
+import java.util.zip.ZipFile;
 import javax.annotation.Nullable;
 import org.apache.commons.io.FilenameUtils;
 import org.gradle.api.Project;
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/java/com/uber/okbuck/core/dependency/DependencyUtils.java
```diff
@@ -122,17 +122,14 @@ static Path getSingleZipFilePath(Project project, File baseDir, String zipToFind
...
-    }
-    return null;
+    return zipFiles
+        .getFiles()
+        .stream()
+        .sorted()
+        .filter(FileUtil::isZipFile)
+        .findFirst()
+        .map(File::toPath)
+        .orElse(null);
```
raviagarwal7: is it possible to maybe get the most recent one?

also, we can probably print a warning for the user that you might see cache misses and should clear these folders to prevent it in future?

kageiit: I would recommend we fix this by making a custom version of the bad artifacts instead of allowing non determinism in the builds

Another option is to make buck not add the source jar to the rulekey

romanoid: we already have this non-determinism depending on which of the source jars given machine fetched. The only thing that we were getting on top is we were getting random failures sometimes in some cases blocking builds and needing immediate attention to proceed.

buildSrc/src/main/java/com/uber/okbuck/core/dependency/DependencyUtils.java
```diff
@@ -122,17 +125,21 @@ static Path getSingleZipFilePath(Project project, File baseDir, String zipToFind
             ImmutableMap.of(
                 "dir", baseDir.getAbsolutePath(), "includes", ImmutableList.of("**/" + zipToFind)));
 
+    return zipFiles
+        .getFiles()
+        .stream()
+        .filter(FileUtil::isZipFile)
+        .min(Comparator.comparing(DependencyUtils::jarComparisonKeyFunction))
```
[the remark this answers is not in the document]
kageiit: nice!

PR #776, opened 2018-12-31
 Add support for new LintRegistry service API in lint detection

hzsweers:
Android lints now should be declared via SPI for IssueRegistry. This adds support for detecting that in okbuck, as it currently still only looks for manifest entries.

Note that since usually one would use `@AutoService`, we can't do this completely automatically. I've added two bits here:

1. Check if there's a service file manually declared
2. A a property to `OkBuckExtension` `containsLintRegistry` to manually declare it.
  * Note I didn't add this to the existing lint extension, as that appears to be solely for configuration of running lints. Lint projects can, themselves, be linted, so I thought it best to leave this separate.


hzsweers:
If I'm reading those build failures right (unfortunately master does not build for me locally), is there not per-project okbuck configuration? If so, should we use a per-project gradle property as a flag instead?

buildSrc/src/main/java/com/uber/okbuck/core/model/jvm/JvmTarget.java
```diff
@@ -18,6 +18,10 @@
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
+
+import kotlin.collections.CollectionsKt;
```
kageiit: Please avoid kotlin stdlib and use standard ja a stream apis instead

libraries/customLintLibrarySpiAuto/build.gradle
```diff
@@ -0,0 +1,16 @@
...
+    compileOnly deps.apt.autoService
+    compileOnly deps.lint.lintApi
+    compileOnly deps.lint.lintChecks
+
+    testImplementation deps.lint.lint
+    testImplementation deps.lint.lintTests
+}
+
+okbuck {
+    containsLintRegistry = true
```
kageiit: This is configured currently at the top level okbuck extension. Let's keep the same behavior for consistency?

buildSrc/src/main/java/com/uber/okbuck/core/model/jvm/JvmTarget.java
hzsweers: Looks like these are just stale imports from when I was testing. Though comment for the future - what I was doing that used this is hardly possible with streams APIs, they're not 1:1

libraries/customLintLibrarySpiAuto/build.gradle
hzsweers: See my comment on the PR: How would we mark a specific module as such if it's only defined all at root level? I'm thinking we should just use a gradle property instead

buildSrc/src/main/java/com/uber/okbuck/core/model/jvm/JvmTarget.java
```diff
@@ -153,11 +154,27 @@ public LintOptions getLintOptions() {
...
         && (jarTask.getManifest().getAttributes().containsKey("Lint-Registry")
-            || jarTask.getManifest().getAttributes().containsKey("Lint-Registry-v2"));
+            || jarTask.getManifest().getAttributes().containsKey("Lint-Registry-v2"))) {
+      // Legacy Lint-Registry approach
+      return true;
+    }
+    if (getMainJavaResourceDirs().stream()
+        .anyMatch(resourceDir -> new File(resourceDir,
+            "META-INF" + File.separator + "services" + File.separator + "com.android.tools.lint.client.api.IssueRegistry")
+            .exists())) {
```
raviagarwal7: nit: use Paths.resolve() api instead of using `File.separator`

buildSrc/src/main/java/com/uber/okbuck/core/model/jvm/JvmTarget.java
```diff
@@ -153,11 +155,24 @@ public LintOptions getLintOptions() {
...
+      // Legacy Lint-Registry approach
+      return true;
+    }
+    // New SPI approach. NOTE this does not cover generated services, such as with AutoService.
+    // Should use the above CONTAINS_LINT_REGISTRY
+    return getMainJavaResourceDirs().stream()
+        .anyMatch(resourceDir -> Paths.get("META-INF/services/com.android.tools.lint.client.api.IssueRegistry")
+            .resolve(resourceDir.toPath())
+            .toFile()
+            .exists());
```
raviagarwal7:
```suggestion
        .anyMatch(resourceDir -> resourceDir.toPath()
            .resolve("META-INF")
            .resolve("services")
            .resolve("com.android.tools.lint.client.api.IssueRegistry")
            .toFile()
            .exists());
```

hzsweers:
I'm not sure I follow the lint failures. It seems almost like it only lints one error per location? Will test more

image

buildSrc/src/main/java/com/uber/okbuck/core/model/jvm/JvmTarget.java
```diff
@@ -159,11 +160,24 @@ public LintOptions getLintOptions() {
...
-            || jarTask.getManifest().getAttributes().containsKey("Lint-Registry-v2"));
+            || jarTask.getManifest().getAttributes().containsKey("Lint-Registry-v2"))) {
+      // Legacy Lint-Registry approach
+      return true;
+    }
+    // New SPI approach. NOTE this does not cover generated services, such as with AutoService.
+    // Should use the above CONTAINS_LINT_REGISTRY
+    return getMainJavaResourceDirs().stream()
+        .anyMatch(resourceDir -> Paths.get("META-INF/services/com.android.tools.lint.client.api.IssueRegistry")
+            .resolve(resourceDir.toPath())
```
raviagarwal7: seems like this change got reverted.

libraries/customLintLibrarySpiAuto/src/main/java/com/uber/lint/ColorDetector.java
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  @Override
+  public void visitAttribute(XmlContext context, Attr attribute) {
+    if (attribute.getValue().contains(SdkConstants.ANDROID_COLOR_RESOURCE_PREFIX)) {
+      context.report(ISSUE, attribute, context.getLocation(attribute),
+          "Using Android color resources are not recommended. Manufacturers are overriding them with other colors.");
+    }
+  }
+}
```
raviagarwal7:
```suggestion
}

```

libraries/customLintLibrarySpi/src/main/java/com/uber/lint/ColorDetector.java
```diff
@@ -0,0 +1,48 @@
...
+  }
+
+  @Override
+  public void visitAttribute(XmlContext context, Attr attribute) {
+    if (attribute.getValue().contains(SdkConstants.ANDROID_COLOR_RESOURCE_PREFIX)) {
+      context.report(ISSUE, attribute, context.getLocation(attribute),
+          "Using Android color resources are not recommended. Manufacturers are overriding them with other colors.");
+    }
+  }
+}
```
raviagarwal7:
```suggestion
}

```

hzsweers:
Talked with @ravi, going to rehash this to support the new lint APIs (including the new `lintChecks` configuration to make consumption declarative rather than the current implicit approach)

PR #775, opened 2018-12-31
Add support for Kotlin -Xfriend-paths

hzsweers:
This adds support for Kotlin's `-Xfriend-paths` option.

Note that I'm not entirely sure this is working, as the project build fails on master for me and there are some parts of okbuck I don't understand, such as how to generate anything into custom rules like `okbuck_android_library` or any other android libraries (they seem to non-configurable?).

Resolves #709. Technically there is an upstream BUCK bug for this as well, but they dismissed it as `wontfix`: https://github.com/facebook/buck/issues/2029. I don't think we need them to implement anything though since it can be done through kotlinc arguments.

Example BUCK rule now:

```python
kotlin_test(
    name = 'test_main',
    srcs = glob([
        'src/test/java/**/*.java',
        'src/test/java/**/*.kt',
    ]),
    source = '8',
    target = '8',
    plugins = [
        '//.okbuck/workspace/processor:processor_com.google.dagger-dagger-compiler-2.16',
    ],
    extra_arguments = [
        '-proc:none',
    ],
    extra_kotlinc_arguments = [
        '-Xplugin=.okbuck/workspace/kotlin_home/libexec/lib/kotlin-allopen.jar',
        '-P',
        'plugin:org.jetbrains.kotlin.allopen:annotation=demo.Anno',
        '-P',
        'plugin:org.jetbrains.kotlin.allopen:annotation=demo.Anno2',
        '-Xfriend-paths=$(location :src_main[output])',
    ],
# ... and the rest
```

Eyeballed https://github.com/JetBrains/kotlin/commit/4bd66d5aa8469ba93b7d354a35f472fefcb7a9e8

PR #777, opened 2018-12-31
Add support for reading custom Kotlin compiler arguments

hzsweers:
This builds on #775 and adds support for reading compiler configuration directly from the kotlin compile task. Not written tests yet because I want to wait for #775 to get in, but it's a start and want to get this up now for early feedback.

This supports all the standard arguments configurable by the gradle plugin, as well as the important free arguments option to add new ones

### DIFF
--- a/buildSrc/src/main/java/com/uber/okbuck/core/model/jvm/JvmTarget.java
+++ b/buildSrc/src/main/java/com/uber/okbuck/core/model/jvm/JvmTarget.java
@@ -1,7 +1,6 @@
 package com.uber.okbuck.core.model.jvm;
 
 import com.android.builder.model.LintOptions;
-import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.uber.okbuck.OkBuckGradlePlugin;
@@ -14,12 +13,14 @@ import com.uber.okbuck.core.model.base.Target;
 import com.uber.okbuck.core.util.ProjectUtil;
 import java.io.File;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.Optional;
 import java.util.ServiceLoader;
 import java.util.Set;
 import java.util.stream.Collectors;
 import javax.annotation.Nullable;
+
 import org.gradle.api.JavaVersion;
 import org.gradle.api.Project;
 import org.gradle.api.Task;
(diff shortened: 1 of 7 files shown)


buildSrc/src/main/java/com/uber/okbuck/core/model/jvm/JvmTarget.java
```diff
@@ -281,6 +285,46 @@ public static String javaVersion(JavaVersion version) {
       return ImmutableList.of();
     }
     ImmutableList.Builder<String> optionBuilder = ImmutableList.builder();
+
+    // Add all compiler arguments
+    String compileTaskName = isTest ? "compileTestKotlin" : "compileKotlin";
```
kageiit: Is there no public API/static constants for these task names? Hardcoding can lead to breakages in future

kageiit: This approach also won't work for kotlin Android targets

hzsweers: No public api/constants. Why won't this work for Android targets?

kageiit: Android targets don't have the same task names as they are variant aware

hzsweers: Good point. I can do something like `tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {`, but how can I connect those back to the specific variant at hand? Or would it just be basic String matching?

hzsweers: alternatively - just grab the first one and only support configurations that apply to all

thalescm: I tend to think people use flavors and buildTypes for feature specific things (changing a hardcoded string value, changing themes, etc...). Could not think of a use case where anyone would want to consistently compile differently depending on that, so I think it's valid to only support configurations that apply to all.

hzsweers: Agreed, just updated to take the first matching one in 245e32f

hzsweers: Did keep the test vs non-test distinction though. Ultimately it's a dynamic naming strategy though

```kotlin
    override val compileKotlinTaskName: String
        get() = lowerCamelCaseName(
            "compile",
            compilationName.takeIf { it != KotlinCompilation.MAIN_COMPILATION_NAME },
            "Kotlin",
            target.targetName
        )
```

hzsweers:
Will do

PR #778, opened 2018-12-31
Add some information about local development to README

hzsweers:
Based on my spelunking through the codebase today

### DIFF
--- a/README.md
+++ b/README.md
@@ -96,7 +96,7 @@ We'd love for you to contribute to our open source projects. Before we can accep
 
 #### Useful commands to get your dev environment up and running
 
-Force rerun of `okbuck`:
+Force rerun of `okbuck`
 
 ```console
 ❯ ./gradlew okbuck -Dokbuck.wrapper=true --stacktrace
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -92,6 +92,27 @@ We'd love for you to contribute to our open source projects. Before we can accep
 - If you **have a feature request**, open an issue or submit an implementation via a pull request
 - If you **want to contribute**, submit a pull request.
 
+### Development
+
```
kageiit: Please move this to contributing.md

hzsweers: Contributing is information to show before filing an issue or PR, which is after local development is started. Most things like this are put into a README section or wiki page

kageiit:
Let's move to a new developing wiki page then. This information is not suitable for readme

hzsweers:
Sure. Do you have any comments on this bit though? (Can't PR wiki changes)

hzsweers:
Sidenote - by that logic, shouldn't the code style section above it also be moved? That was the main reason I figured this project followed conventions with having local development info in the README

PR #794, opened 2019-01-08
Add a test case for AndroidManifest merge error

raviagarwal7:
google issue tracker - https://issuetracker.google.com/issues/114358542


raviagarwal7:
should be fixed when https://github.com/facebook/buck/pull/2153 is landed on Buck.

PR #800, opened 2019-01-18
Created a shared ui library 

jamescoggan:
There is an issue when using shared drawables from modules
Ok buck does not seem to share them across the modules
When trying to compile the kotlin-app the build fails with:

```
kotlin-app/src/main/java/com/uber/okbuck/example/MainActivity.kt:24:35: error: unresolved reference: icon_smile
        smile = findViewById(R.id.icon_smile)
```


kotlin-app/src/main/java/com/uber/okbuck/example/MainActivity.kt
```diff
@@ -12,11 +14,16 @@ class MainActivity : DaggerAppCompatActivity() {
...
     lateinit var analytics: Analytics
 
+    lateinit var smile : ImageView
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_main)
         setSupportActionBar(toolbar)
 
+        smile = findViewById(R.id.icon_smile)
```
artem-zinnatullin: Can you try to reference it from Java class in this module? I think I saw exact same problem but it only happened with Kotlin classes

thalescm: I added the same thing in `app` project (which is java) and it fails the same way.
Trying to find the root cause here

thalescm:
So, in terms of okubck itself, there is no issue. You can check that on `kotlin-app/BUCK` file, `res_debug` rule depends on `sharedui:res_debug`, and that `src_debug` depends on both `sharedui:res_debug` and it's own `res_debug`.
This means that at compilation time it have access to everything it's supposed to, and okbuck generated the BUCK files correctly.

The main issue is inside buck itself I think. For some reason when you try to find `R.id.icon_smile`, it will try to find on `com.uber.okbuck.example.R`, which is obviously not there. Gradle somehow is able to treat it differently, or provides a mechanism to search for dependencies R files as well.

If you add this in kotlin-app's MainActivity:
`import com.uber.okbuck.example.sharedui.R as R2`
and then do:
`smile = findViewById(R2.id.icon_smile)`

it compiles fine

thalescm:
I'm closing the PR since it's non related to okbuck :)

PR #804, opened 2019-01-24
Fix rules overrides bug

### LINKED ISSUE #803: Add ruleType for okbuck_manifest .raw
This will enable external users of okbuck to overrite that rule if needed

thalescm:
Fixes bug which prevent okbuck_keystore and okbuck_manifest rules to  be overriden

Resolves #803

### DIFF
--- a/buildSrc/src/main/rocker/com/uber/okbuck/template/android/KeystoreRule.rocker.raw
+++ b/buildSrc/src/main/rocker/com/uber/okbuck/template/android/KeystoreRule.rocker.raw
@@ -4,9 +4,9 @@ String storePassword,
 String keyAlias,
 String keyPassword
 )
-@com.uber.okbuck.template.base.BuckRule.template() -> {
+@(ruleType)(
     storeFile = '@storeFile',
     storePassword = '@storePassword',
     keyAlias = '@keyAlias',
     keyPassword = '@keyPassword',
-}
+)
--- a/buildSrc/src/main/rocker/com/uber/okbuck/template/android/KeystoreRule.rocker.raw
+++ b/buildSrc/src/main/rocker/com/uber/okbuck/template/android/KeystoreRule.rocker.raw
@@ -5,10 +5,10 @@ String storePassword,
 String keyAlias,
 String keyPassword
 )
-okbuck_keystore(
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/rocker/com/uber/okbuck/template/android/KeystoreRule.rocker.raw
```diff
@@ -1,14 +1,12 @@
...
 String keyPassword
 )
-okbuck_keystore(
-  name = '@name',
-  storeFile = '@storeFile',
-  storePassword = '@storePassword',
-  keyAlias = '@keyAlias',
-  keyPassword = '@keyPassword'
-)
+@com.uber.okbuck.template.base.BuckRule.template() -> {
```
romanoid: technically you don't need to inherit from the template, instead you may just use `@(ruleType)` instead of `okbuck_keystore` 

this may be cleaner since it is less likely to introduce additional problems

buildSrc/src/main/rocker/com/uber/okbuck/template/android/ManifestRule.rocker.raw
```diff
@@ -9,8 +9,7 @@ String versionName,
 boolean debuggable,
 Collection secondaryManifests
 )
-okbuck_manifest(
-    name = '@name',
+@com.uber.okbuck.template.base.BuckRule.template() -> {
```
romanoid: same as okbuck_keystore here. 

If you want to proceed with these changes, you'll have to carefully check rulekey changes on the repo that uses OkBuck to verify there are no unexpected changes.

buildSrc/src/main/rocker/com/uber/okbuck/template/config/OkbuckTargets.rocker.raw
```diff
@@ -242,6 +243,7 @@ def okbuck_manifest(
         out = 'AndroidManifest.xml',
         type = 'okbuck_manifest',
         cmd = ' '.join(cmds),
+        visibility = visibility,
```
romanoid: this rule was not previously public and now it is. Maybe it is better to keep it private?

thalescm:
@romanoid done :D

PR #805, opened 2019-01-26
Infer sources using ResolvedArtifactResult instead of doing filesystem check

raviagarwal7:


buildSrc/src/main/java/com/uber/okbuck/core/dependency/ExternalDependency.java
```diff
@@ -92,6 +89,16 @@ public File getRealDependencyFile() {
     return this.base.realDependencyFile();
   }
 
+  /** Gets the real sources jar file for the dependency. */
+  public Optional<File> getRealSourceFile() {
+    // Doesn't support sources for artifacts with classifier
```
kageiit: do we need to create a follow up for this?

PR #814, opened 2019-02-11
Adding arm64-v8a as a supported ABI to okbuck

naveenOnarayanan:
* Simply adding support to `arm64-v8a` as a recognizable attribute for ABI in okbuck

Pulling in the `ARM64` enum from - https://github.com/facebook/buck/blob/a6c7d7391a7a8e575c6372cac6690a240b5907ea/src/com/facebook/buck/android/toolchain/ndk/TargetCpuType.java


kageiit:
Can you please rebase?

Issue #817, opened 2019-02-13
Fix broken -Xfriend-paths after kotlin 1.3.20 upgrade


kageiit:
See https://github.com/uber/okbuck/pull/802#issuecomment-463326871

thalescm:
Apparently it should working wither by passing the directories with the classes, or by passing the archive, as [tested here](https://github.com/JetBrains/kotlin/blob/v1.3.20/compiler/tests/org/jetbrains/kotlin/cli/FriendPathsTest.kt).

Looking at the stacktrace, when reverting [this](https://github.com/uber/okbuck/pull/802/files#diff-2c2c5cba1a8a3ec4aab0fa722e1d0ad8L3) (so we get the error), and running
`./buckw test //libraries/kotlinlibrary:test_main -v 10`
we can see that friend paths is set:
```
-Xfriend-paths=buck-out/gen/libraries/kotlinlibrary/lib__src_main__output/src_main.jar
```

Looking at the generated BUCK file (on `libraries/kotlinlibrary/BUCK`:
```
kotlin_test(
    name = 'test_main',
    <...>,
    extra_kotlinc_arguments = [
        <...>
        '-Xfriend-paths=buck-out/gen/libraries/kotlinlibrary/lib__src_main__output/src_main.jar',
    ],
    <...>,
)
```
It seems that kotlin compiler is not able to use relative paths for `-Xfriend-paths` option. Checked gradle output, it uses absolute path on this argument. And if I change the BUCK file to use the absolute path to that jar, tests will compile as well.

thalescm:
https://youtrack.jetbrains.com/issue/KT-29933

kageiit:
Good catch. Another option we have in buck is to expand the path ourselves in the buck kotlin rules

PR #821, opened 2019-02-15
Add `friend_paths` to kotlin rules

thalescm:
Also added a custom way to define custom parameter for rules other than just `extra_arguments` (which was the only possible difference until now).
This is extremely important on okbuck support to kotlin as there are plans to add a better API to kotlin rules in the future.

**Depends on** https://github.com/facebook/buck/pull/2197
Resolves https://github.com/uber/okbuck/issues/817


kageiit:
Its weird that friend paths are the only oddball case where they cannot fit in custom arguments

ZacSweers:
> where they cannot fit in custom arguments

They can, it's just something that should be implicitly handled by the build system rather than be something the consumer needs to implement. They're a compiler arg like anything else.

I'm not sure I understand the goal of this PR - is it to remove dependence on explicitly declaring `extra_kotlinc_args` in BUCK (since that PR to standardize configuration has gone nowhere?)

thalescm:
> is it to remove dependence on explicitly declaring `extra_kotlinc_args` in BUCK
Yes, following what I said [here](https://github.com/facebook/buck/pull/2030#pullrequestreview-169689266)

The idea is to expand kotlin rule's API to be more user friendly, with proper documentation instead of having the user need to dig into the compiler to see what he needs. To better extend the compiler.

As for:
> it's just something that should be implicitly handled by the build system rather than be something the consumer needs to implement  

I agree, but buck doesn't necessarily know what a test rule is testing to automatically add this, unless the user explicitly configure the "source" rule to add the test one on `tests` param. So it would be possible to change the reproductiveness of a test rule just by adding a parameter on another rule, which goes against buck's vision.

And last:
> Its weird that friend paths are the only oddball case where they cannot fit in custom arguments

Custom arguments is a concept the just emerged inside okbuck. As before the only custom config was directly coupled to `extra` arguments to a compiler (being it java, kotlin or scala).
Since there is this plan to add more APIs to kt, this PR also created the concept of custom configs so that each compiler can determine what they need in an easy way.

kageiit:
@thalescm buck version is updated to latest with friend paths fix included. Please fix the ci issues and merge when green

Issue #828, opened 2019-03-17
Support for multiDexEnabled in buildTypes


bangarharshit:
Verified that `gradle` supports multidexEnabled in buildTypes as well. 
```
buildTypes {
    release {
        multiDexEnabled false
    }
    debug {
        multiDexEnabled true
    }
}
```

`okbuck` is ignoring the same [code](https://github.com/uber/okbuck/blob/e70704434418705f467dded57bc5a15e8f48c80f/buildSrc/src/main/java/com/uber/okbuck/core/model/android/AndroidAppTarget.java#L55). We should add a check based on `getBaseVariant().getBuildType().getMultiDexEnabled()`.

kageiit:
Merged #829

PR #829, opened 2019-03-17
[Demo diff] Added support for reading multiDexEnabled from build type based config

bangarharshit:
Demo diff for https://github.com/uber/okbuck/issues/828. It first checks whether `multiDexEnabled` is defined in `buildType`. If not present, then check  in `defaultConfig`.

### DIFF
--- a/buildSrc/src/main/java/com/uber/okbuck/core/model/android/AndroidAppTarget.java
+++ b/buildSrc/src/main/java/com/uber/okbuck/core/model/android/AndroidAppTarget.java
@@ -52,12 +52,9 @@ public class AndroidAppTarget extends AndroidLibTarget {
     Set<String> filters = ndkCompile != null ? ndkCompile.getAbiFilters() : ImmutableSet.of();
     cpuFilters = filters != null ? filters : ImmutableSet.of();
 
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/java/com/uber/okbuck/core/model/android/AndroidAppTarget.java
```diff
@@ -52,12 +52,9 @@ public AndroidAppTarget(Project project, String name, boolean isTest) {
...
 
-    Boolean multidex = getBaseVariant().getMergedFlavor().getMultiDexEnabled();
-    if (multidex == null) {
-      multidexEnabled = false;
-    } else {
-      multidexEnabled = multidex;
-    }
+    multidexEnabled = Optional.ofNullable(getBaseVariant().getBuildType().getMultiDexEnabled())
+            .orElse(Optional.ofNullable(getBaseVariant().getMergedFlavor().getMultiDexEnabled())
+                    .orElse(false));
```
raviagarwal7: seems like the mergedFlavour doesn't respect the multiDexEnabled value on the buildType. 

cc: @kageiit

Rada8732 (approved):

Issue #797, opened 2019-01-15
Error building (mvn artifact hash mismatch) after successful OkBuck setup


erawhctim:
I setup okbuck properly (to my knowledge) and now I'm getting a build error:
`Unable to download mvn:com.google.j2objc:j2objc-annotations:jar:1.1 (hashes do not match. Expected 2994a7eb78f2710bd3d3bfb639b2c94e219cedac0d4d084d516e78c16dddecf6, saw 40ceb7157feb263949e0f503fe5f71689333a621021aa20ce0d0acee3badaa0f)`

What does that mean? 

build command: `./buckw build //app:bin_stagingDebug`

Looking at the dependency graph via `./gradlew dependencies`, it looks like the j2ObjC lib is required by Truth, which is setup as a test-only dependency (if that's helpful).

I'm completely new to Buck/okbuck, so let me know if there is more detail I can provide or if this makes more sense as an issue on the main Buck repo (or a StackOverflow question). Just looking to gain insight on what this error means to help me investigate.

kageiit:
Which version of okbuck, gradle and android gradle plugin are you using?

erawhctim:
okbuck: 0.46.2
gradle: 4.7
gradle android plugin: 3.2.1

If I manually edit the SHA in `.okbuck/ext/com/google/j2objc/BUCK` to the hash that buck expects, the build succeeds - I don't yet know if that's ok or problematic for us, but it seems to solve my problem, so at this point I'm just more curious as to why it happened/where the discrepancy came from.

kageiit:
@raviagarwal7 has been looking at a similar issue in https://github.com/uber/okbuck/pull/796

This could be the same issue

erawhctim:
Good to know, thanks! I'm on Kotlin v1.3.10 as well, could very well be the same thing.

erawhctim:
@raviagarwal7 ok great - thanks so much!

raviagarwal7:
Seems like this is probably a metadata caching issue b/w different gradle versions. We might want to wipe out the gradle cache completely in such scenario to make sure the gradle cache is consistent.

an example with `kotlin-allopen`

With existing gradle home
```
╰─$ find /Users/ravi/.gradle/caches/modules-2/files-2.1/org.jetbrains.kotlin/kotlin-allopen/1.3.11                                                                                                                                      130 ↵
[...]

Issue #322, opened 2017-01-11
Configuration with name 'buckLint_deps' not found.


hai-nguyen:
Hi,

We got this message when trying to run `okbuck` task: `Configuration with name 'buckLint_deps' not found.`

Is there any way I can skip the lint check step? In our case, we need to skip Lint check due to some 3rd libraries we are using.

Thanks.

kageiit:
You can disable lint entirety using

```
okbuck {
  lint {
    disabled = true
  }
}
```

pchmielowski:
@kageiit Where should I put this code?

runningcode:
inside the okbuck block in your root build.gradle. check the okbuck repo for a sample.

BingranLi:
@kageiit I also got the same error. Is there a way to fix it without disabling lint?

kageiit:
okbuck.lint.version is not mandatory. The default version is inferred from the Android Gradle plugin version

kageiit:
The default version is set during okbuck initialization here: https://github.com/uber/okbuck/blob/master/buildSrc/src/main/java/com/uber/okbuck/extension/LintExtension.java#L24 so that is used if no user supplied value exists

TonyTangAndroid:
For anyone who is wondering the compilable code block to make it work without disabling the link check, please refer to this example 

```
okbuck {
    lint {
        version = '26.3.2'
    }
}
```
[26.3.2](https://mvnrepository.com/artifact/com.android.tools.lint/lint/26.3.2) is the latest version of `com.android.tools.lint`.

NickFirmani:
Hey folks, I think that this could use some more investigation, the two options to work around this right now are either disabling lint or forcing it to use a specific version, which then breaks `//...`

I've created a sample project here: https://github.com/NickFirmani/android-template/tree/f5ec67d14042159e907b1b9375e2bcecea88ac56 that reproduces this problem hermetically

kageiit:
Thanks for providing a sample @NickFirmani . We will take a look

stargt:
i've encountered same problem.

```
* What went wrong:
Execution failed for task ':MODULE_NAME:okbuck'.
> Configuration with name 'buckLint_deps' not found.
```

How can I solve it?

kageiit:
Should be fixed in latest version

PR #830, opened 2019-03-21
Update instructions for listing all buck targets in README

TonyTangAndroid:
It takes a while to make a HelloWorld android project work with okbuck due to two issues
1, It needs to disable the android lint to avoid issue https://github.com/uber/okbuck/issues/322
2, `./buckw targets` is broken and we have to execute `./buckw targets //...` due to issue https://github.com/facebook/buck/issues/1930.
3, issue https://github.com/uber/okbuck/issues/797 also stops the okbuck to work in new project.

With this pull request, it will resolve these there issue all together and make okbuck ok again.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,15 +30,6 @@ repositories {
 }
 
 apply plugin: 'com.uber.okbuck'
-okbuck {
-
-    lint {
-        disabled = true
-    }
-    externalDependencies {
-        downloadInBuck = false
-    }
-}
 
 ```
 
--- a/README.md
+++ b/README.md
@@ -30,7 +30,6 @@ repositories {
 }
 
 apply plugin: 'com.uber.okbuck'
-
 okbuck {
 
     lint {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -30,6 +30,17 @@ repositories {
 }
 
 apply plugin: 'com.uber.okbuck'
+
```
kageiit: This should not be needed by default. Please remove this blockk

README.md
```diff
@@ -55,7 +66,7 @@ Please make sure you have watchman installed for `buckw` to run `okbuck` only wh
 
 ```bash
 # List all buck targets
-./buckw targets
+./buckw targets //...
```
kageiit: This is valid so should be fine to merge

README.md
TonyTangAndroid: I have removed the block. I was under impression that it would throw error and stop buck work if I remove it. I will check it in next pull request and provide more context by creating an issue if I reproduce it. For now, let' get the command line updated.

PR #838, opened 2019-04-17
Cleanup versionless resolution actions handling

raviagarwal7:
This allows for only the below resolution actions:

- latest
- all
- single (fails if finds multiple versions)

When using latest or single versionless names will be used for external dependencies.

Can also specify coordinates for deps whose all versions is to be allowed when in latest or single mode. for eg: roblectric runtime deps.

```
externalDependency {
    resolutionAction = latest
    allowAllVersions = [
        "org.robolectric:android-all",
        "com.guava:guava",
    ]
}
```

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -211,35 +211,7 @@ okbuck {
     }
 
     externalDependencies {
-        versionless = true
-        allowLatestVersion = [
-            "com.google.auto.value:auto-value",
-            "com.google.code.findbugs:jsr305",
-            "org.ow2.asm:asm",
-            "org.ow2.asm:asm-commons",
-            "com.google.protobuf:protobuf-java",
-            "commons-codec:commons-codec",
-            "com.google.errorprone:error_prone_annotations",
-            "com.google.auto:auto-common",
-            "com.google.guava:guava",
-            "org.apache.httpcomponents:httpcore",
-            "org.jetbrains.kotlin:kotlin-stdlib",
-            "org.jetbrains.kotlin:kotlin-reflect",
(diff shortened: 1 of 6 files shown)


buildSrc/src/main/java/com/uber/okbuck/extension/ResolutionAction.java
```diff
@@ -0,0 +1,18 @@
...
+  // Use all resolved versions of the external dependency
+  ALL("all"),
+
+  // Use single resolved version of the external dependency and fail if multiple versions are found
+  SINGLE("single");
+
+  private final String action;
+
+  ResolutionAction(String action) {
+    this.action = action;
```
kageiit: why do we need action?

Its the same as enum.name().toLowerCase()

build.gradle
```diff
@@ -211,35 +211,7 @@ okbuck {
...
-            "com.squareup:javapoet",
-            "org.checkerframework:checker-compat-qual",
-            "com.google.code.gson:gson",
-            "org.apache.commons:commons-compress",
-            "org.apache.httpcomponents:httpmime",
-        ]
-        allowAllVersions = [
-            "org.robolectric:android-all",
-        ]
+        resolutionAction = "latest"
```
thalescm: I like when we have as many examples as possible in here for people to refer to. So I'd suggest to add an `allowAllVersions` list here.

buildSrc/src/main/java/com/uber/okbuck/extension/ResolutionAction.java
raviagarwal7: removed it

PR #846, opened 2019-05-22
Do not check for lint manifest attributes

kageiit:
**Description**:
Custom lint targets can use AutoService and generate a compatible jar with an issue registry so we should not gate on having tthe attribute available in the gradle file also


raviagarwal7:
I think this was needed for lint projects in the repo since okbuck is run before autoservice could generate the issue registry.

kageiit:
Its actually not needed if okbuck just adds projects from `buckLint` config to the custom lint option as a macro

PR #851, opened 2019-06-04
Android modules rule

thalescm:
Creates `okbuck_android_module` and `okbuck_kotlin_android_module` rules that represents the full description of an android module in buck, thus simplifying manual creation of rules, if necessary.


buildSrc/src/main/rocker/com/uber/okbuck/template/config/OkbuckAndroidModules.rocker.raw
```diff
@@ -0,0 +1,113 @@
...
+
+    src_kwargs.update({'deps' : src_deps})
+    if src_exported_deps:
+        src_kwargs.update({'exported_deps' : src_exported_deps})
+    if language:
+        src_kwargs.update({'language' : language})
+
+    @(androidLibraryRule)(**src_kwargs)
+
+def okbuck_kotlin_android_module(
```
raviagarwal7: we should allow overriding `okbuck_kotlin_android_module` & `okbuck_android_module` as well.

This way we can override them to be `uber_kotlin_android_module` & `uber_android_module` in the android repo. These override can then directly call respective `okbuck_*` impl.

buildSrc/src/main/rocker/com/uber/okbuck/template/config/OkbuckAndroidModules.rocker.raw
```diff
@@ -0,0 +1,113 @@
...
+        visibility = None,
+        package=None,
+        ## Manifest related args
+        manifest_main_manifest = None,
+        manifest_min_sdk = None,
+        manifest_target_sdk = None,
+        manifest_version_code=None,
+        manifest_version_name=None,
+        manifest_debuggable=False,
+        manifest_secondary_manifests=None,
```
raviagarwal7: could we make this a struct? https://docs.bazel.build/versions/master/skylark/lib/struct.html

one for manifest & one for resources. would be cleaner I think.

raviagarwal7: nvm, that would boil down to the same thing as before :|

buildSrc/src/main/java/com/uber/okbuck/generator/BuckFileGenerator.java
```diff
@@ -210,8 +204,8 @@ public static void generate(
 
     libRules.forEach(
         rule -> {
-          if (rule instanceof ResourceRule && rule.name() != null) {
-            deps.add(rule.buckName());
+          if (rule instanceof AndroidModuleRule && rule.name() != null) {
+            deps.add(rule.buckName().replace(":src_", ":res_"));
```
raviagarwal7: should this be moved to exopackage's macro?

thalescm: these `deps` relates to `android_binary` (normal one or binary that correspond to a library instrumentaion apk) rule, not exopackage.
We could do that, but I didn't want to make a breaking change to whoever already override it today (forcing them to modify their defs)

buildSrc/src/main/rocker/com/uber/okbuck/template/config/OkbuckAndroidModules.rocker.raw
thalescm: This is already allowed.
Although it's kinda of strange to declare, since you'd have to call the `nativeRuleName` as `android_module`.
I'll update the PR with the example so people can refer to

PR #852, opened 2019-06-05
Small fix on android module macro

thalescm:
Basically:
* res_release should go into the deps of the library and exported_deps of the corresponding resource (if respective src_release is in exported deps)
* res_release should go into the deps of the library and deps of the corresponding resource (if not respective src_release not exported)

This PR fixes the part of having the `res_release` as dep, in case it respective `src` was exported

EDIT

This PR also fixes missing `res` rules as dependencies for test rules (robolectric and instrumentation)

### DIFF
--- a/buildSrc/src/main/rocker/com/uber/okbuck/template/config/OkbuckAndroidModules.rocker.raw
+++ b/buildSrc/src/main/rocker/com/uber/okbuck/template/config/OkbuckAndroidModules.rocker.raw
@@ -93,7 +93,10 @@ def okbuck_android_module(
     )
     src_kwargs.update(kwargs)
 
-    src_deps = deps + res_lib_deps + [get_rule_dep(res_rule_name), get_rule_dep(manifest_rule_name)]
+    src_deps = deps + \
+               res_lib_deps + \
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/rocker/com/uber/okbuck/template/config/OkbuckAndroidModules.rocker.raw
```diff
@@ -93,7 +93,10 @@ def okbuck_android_module(
     )
     src_kwargs.update(kwargs)
 
-    src_deps = deps + res_lib_deps + [get_rule_dep(res_rule_name), get_rule_dep(manifest_rule_name)]
+    src_deps = deps + \
+               res_lib_deps + \
+               res_exported_lib_deps + \
+               [get_rule_dep(res_rule_name), get_rule_dep(manifest_rule_name)]
```
raviagarwal7: change it to below instead.
```
    src_deps = (deps +
               res_lib_deps +
               res_exported_lib_deps +
               [get_rule_dep(res_rule_name), get_rule_dep(manifest_rule_name)])
```

PR #862, opened 2019-07-31
Ensure that test env dict is printed in sorted manner.

raviagarwal7:
Verified:

```
    env = {
        'CACHE_CASSANDRA_CONTACT_POINTS': '[testhost]',
        'CACHE_CASSANDRA_PORT': '1234',
        'OKBUCK_DATACENTER': 'testdc1',
        'OKBUCK_PORT_HTTP': '4321',
        'TEST_ENV': 'true',
    },
```

### DIFF
--- a/buildSrc/src/main/java/com/uber/okbuck/core/model/jvm/TestOptions.java
+++ b/buildSrc/src/main/java/com/uber/okbuck/core/model/jvm/TestOptions.java
@@ -2,22 +2,23 @@ package com.uber.okbuck.core.model.jvm;
 
 import java.util.List;
 import java.util.Map;
+import java.util.TreeMap;
 
 public final class TestOptions {
 
   private final List<String> jvmArgs;
-  private final Map<String, Object> env;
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/java/com/uber/okbuck/core/model/jvm/TestOptions.java
```diff
@@ -2,22 +2,23 @@
...
-    this.env = env;
+    this.env = new TreeMap<>(env);
   }
 
   public List<String> getJvmArgs() {
     return jvmArgs;
   }
 
-  public Map<String, Object> getEnv() {
+  public TreeMap<String, Object> getEnv() {
```
rzsombor: Do we need to surface the underlying type on this getter? Looks like it could remain a Map.

raviagarwal7: true, I will change that.

PR #867, opened 2019-08-09
Don't populate resources or sources in jvm rules if it doesn't contain any valid file.

raviagarwal7:


buildSrc/src/main/java/com/uber/okbuck/core/util/FileUtil.java
```diff
@@ -53,7 +56,26 @@ public static void copyResourceToProject(String resource, File destination) {
   public static ImmutableSet<String> available(Project project, Collection<File> files) {
     return files
         .stream()
-        .filter(File::exists)
+        .filter(
+            rootFile -> {
+              if (rootFile.isDirectory() && rootFile.exists()) {
+                // Check if the directory contains any valid file
+                try (Stream<Path> fileTree = Files.walk(rootFile.toPath())) {
```
thalescm: why is this necessary? what does `Collection<File> file` parameter represent then? I'd imagine, by looking at the method, that they would be all the files in the project

raviagarwal7: This is needed to prevent OkBuck from creating resources arguments in the buck file. 

If buck files are checked in, resources entries will be created on local machine but maybe not on a different machine which doesn't have the folder. We need to add resource entries only if the folder has some valid file. 
Since git ignores folders this is happening quite a lot.

PR #870, opened 2019-08-20
Pad version names so buck sorts them properly

raviagarwal7:


buildSrc/src/main/java/com/uber/okbuck/core/dependency/BaseExternalDependency.java
```diff
@@ -84,7 +87,11 @@ public String packaging() {
   public String targetName() {
     StringBuilder targetName = new StringBuilder(versionless().name());
     if (isVersioned()) {
```
raviagarwal7: add a new method `boolean usePaddedVersion()` which gets initialized based on the external dependency extension setting.

buildSrc/src/main/java/com/uber/okbuck/core/dependency/BaseExternalDependency.java
```diff
@@ -84,7 +87,11 @@ public String packaging() {
...
-      targetName.append(NAME_DELIMITER).append(version());
+      String paddedOVersion = Arrays.stream(version().split("\\."))
```
raviagarwal7: add a new method `paddedVersion()` which is used when padding is enabled otherwise uses the normal `version()`

PR #871, opened 2019-08-21
Integration tests support

thalescm:
**Description**:
Added support for integration tests on java modules.

In order to enable it for your project, there are a few steps required:
1. Set `enableIntegrationTests` on `test` extension in okbuck
```groovy
okbuck {
    <... other configs ... >
    test {
        enableIntegrationTests = true
    }
}
```
2. Copy the file under `tooling/gradle/integration_tests.gradle` (of this project) to your own project
The file contents should have, at minimal, this format:
```groovy
sourceSets {
    integrationTest
}

// define custom test task for running integration tests
// This is used by okbuck in order to determine vm args and env variables
task integrationTest(type: Test) {
    description = 'Runs the integration tests.'
    group = 'verification'
    testClassesDirs = sourceSets.integrationTest.output.classesDirs
    classpath = sourceSets.integrationTest.runtimeClasspath
    outputs.upToDateWhen { false }
}
```
3. Apply file described at [1] on your module to create integration tests, right after applying your language plugin.
```groovy
apply plugin: "java-library"
apply from: rootProject.file("tooling/gradle/integration-tests.gradle")

<... rest of your configurations as they were before ...>
```
4. Now you can define the properties for `integrationTest` similar of how you'd define for `test`.
```
apply plugin: "java-library"
apply from: rootProject.file("tooling/gradle/integration-tests.gradle")

test {
    environment "TEST_ENV", "true"
    environment "CACHE_CASSANDRA_PORT", "1234"
    environment "CACHE_CASSANDRA_CONTACT_POINTS", "[testhost]"
    environment "OKBUCK_DATACENTER", "testdc1"
    environment "OKBUCK_PORT_HTTP", "4321"
}

integrationTest {
    environment "ANOTHER", "true"
}

dependencies {
    api deps.external.gson
    implementation deps.external.dagger

    annotationProcessor deps.apt.javax
    annotationProcessor deps.apt.daggerCompiler

    compileOnly deps.androidx.annotations

    testImplementation deps.test.junit

    integrationTestImplementation deps.test.junit
    integrationTestAnnotationProcessor deps.apt.daggerCompiler
    integrationTestImplementation deps.external.avroIpc
}
```

**Caveats**
Ideally we would want to not have steps two and three. That would require us to be able to map the configuration under that file into calling the correct APIs inside the plugin and adding them programatically there, jumping directly for step 4. I could not find an easy/straight forward way to do that, so I'm open for suggestions.


buildSrc/src/main/java/com/uber/okbuck/generator/BuckFileGenerator.java
```diff
@@ -45,12 +47,19 @@ private BuckFileGenerator() {}
...
+      Project project,
+      BuckFileManager buckFileManager,
+      VisibilityExtension visibilityExtension,
+      TestExtension testExtension) {
     File moduleDir = project.getBuildFile().getParentFile();
     File visibilityFile = new File(moduleDir, visibilityExtension.visibilityFileName);
     boolean hasVisibilityFile = visibilityFile.isFile();
+
+    File integrationTestsFile = new File(moduleDir, testExtension.integrationTestDir);
+    boolean hasIntegrationTests = testExtension.enableIntegrationTests && integrationTestsFile.isDirectory();
```
raviagarwal7: this should be `target.getIntegrationTest().getSources().isEmpty()`

I think the scope might just return an empty configuration scope since the `integrationTest` configuration is not resolvable.

build.gradle
```diff
@@ -180,6 +180,7 @@ okbuck {
         espressoForLibraries = true
         robolectric = true
         robolectricApis = ["21", "27"]
+        enableIntegrationTests = true
```
raviagarwal7: probably name it `integrationTestForJava`

buildSrc/src/main/java/com/uber/okbuck/generator/BuckFileGenerator.java
thalescm: This fails for java targets without the proper tasks. The way you described was my first attempt.
If we solve the way to automatically add the sourceSets and integrationTest task for all jvm targets on okbuck itself, this check is not needed anymore

build.gradle
thalescm: The current work is for all jvm targets, so it should work out of the box. I've tested with kotlin, for example, and was able to setup integration tests for it as well

buildSrc/src/main/java/com/uber/okbuck/composer/jvm/JvmBuckRuleComposer.java
```diff
@@ -21,6 +21,10 @@ public static String test(JvmTarget target) {
     return "test_" + target.getName();
   }
 
+  public static String integrationTest(JvmTarget target) {
```
edsilfer: The name of this method can be made better? What does this string represents? is that the target name? If so it could be `getIntegrationTestTargetName`

buildSrc/src/main/java/com/uber/okbuck/composer/jvm/JvmIntegrationTestRuleComposer.java
```diff
@@ -0,0 +1,67 @@
...
+import com.uber.okbuck.core.model.base.RuleType;
+import com.uber.okbuck.core.model.base.SourceSetType;
+import com.uber.okbuck.core.model.jvm.JvmTarget;
+import com.uber.okbuck.template.core.Rule;
+import com.uber.okbuck.template.jvm.JvmRule;
+
+import java.util.List;
+import java.util.Set;
+
+public final class JvmIntegrationTestRuleComposer extends JvmBuckRuleComposer {
```
edsilfer: Missing JavaDocs?

buildSrc/src/main/java/com/uber/okbuck/core/task/IntegrationTestTask.java
```diff
@@ -0,0 +1,7 @@
+package com.uber.okbuck.core.task;
+
+import org.gradle.api.tasks.testing.Test;
+
+public class IntegrationTestTask extends Test {
```
edsilfer: Can you add some documentation on why this class exists and is empty?

buildSrc/src/main/java/com/uber/okbuck/core/model/base/TargetCache.java
```diff
@@ -52,7 +52,13 @@ public TargetCache(Project project) {
         case KOTLIN_LIB:
           targets =
               ImmutableMap.of(
-                  JvmTarget.MAIN, new JvmTarget(project, JvmTarget.MAIN, "kapt", "kaptTest"));
+                  JvmTarget.MAIN,
+                  new JvmTarget(
+                      project,
+                      JvmTarget.MAIN,
+                      "kapt",
```
edsilfer: Should we have these kapt string in an Enum instead of hard coding them?

tooling/gradle/integration_tests.gradle
```diff
@@ -0,0 +1,13 @@
...
+
+// define custom test task for running integration tests
+// This is used by okbuck in order to determine vm args and env variables
+task integrationTest(type: Test) {
+    description = 'Runs the integration tests.'
+    group = 'verification'
+    testClassesDirs = sourceSets.integrationTest.output.classesDirs
+    classpath = sourceSets.integrationTest.runtimeClasspath
+    outputs.upToDateWhen { false }
+}
```
raviagarwal7: since this can be reused by modules, we should move it into `.okbuck/gradle` generated folder so users can then include it directly rather than having to define one.

tooling/gradle/integration_tests.gradle
```diff
@@ -0,0 +1,13 @@
+sourceSets {
```
raviagarwal7: nit: rename to `integration-test.gradle`

buildSrc/src/main/java/com/uber/okbuck/extension/TestExtension.java
```diff
@@ -18,4 +18,7 @@
 
   /** Enable generation of espresso test rules on android libraries. */
   public boolean espressoForLibraries = false;
+
+  /** Enable creation of integration-test rules for modules with `integration-test` folder */
```
raviagarwal7: I think it will default to `integrationTest` now? please also update the comment.

buildSrc/src/main/java/com/uber/okbuck/composer/jvm/JvmBuckRuleComposer.java
thalescm: This follow the conventions for all other methods (such as `src(`, `bin(` and `test(`. Refactoring it it out of the scope of this PR

buildSrc/src/main/java/com/uber/okbuck/composer/jvm/JvmIntegrationTestRuleComposer.java
thalescm: None of the composers got JavaDocs, I personally think that adding a java doc that would say `Class responsible for composing Jvm IntegrationTest Rules` doesn't add any information that the name itself already says

buildSrc/src/main/java/com/uber/okbuck/core/task/IntegrationTestTask.java
thalescm: Nice catch, this should be deleted

buildSrc/src/main/java/com/uber/okbuck/core/model/base/TargetCache.java
thalescm: You're right, but probably out of the scope of this PR to replace all occurrences of KAPT for the String

buildSrc/src/main/java/com/uber/okbuck/composer/jvm/JvmIntegrationTestRuleComposer.java
edsilfer: For industry convention I personally think that the name should be `Factory` then, since the class is responsible for initialising objects. When first reading `Compose` one might assume that it does more than a `Factory`  since it was not named that way. But again it would fall off the scope of this PR

tooling/gradle/integration_tests.gradle
raviagarwal7: On second thought I think it will bring some cyclic dependency issues. if the file is not there okbuck could fail if a project is applying this file. 

we can leave as is and ask users to create a file and apply from there

buildSrc/src/main/java/com/uber/okbuck/composer/jvm/JvmIntegrationTestRuleComposer.java
thalescm: Agreed

Issue #873, opened 2019-08-28
okbuck fails with Gradle 5.1.1+ (Could not create symlink: FileAlreadyExistsException)


anoopc1:
### Exception Details ([complete-stack-trace](https://gist.github.com/anoopc1/0f728f696f117f3058ae1b3dbc959b83))
```
Could not create symlink /<repo>/.okbuck/ext/<module-path>/data-collector-3.4.0-sources.jar -> /<home>/.gradle/caches/modules-2/files-2.1/<module-path>/3.4.0/6f4575ecbe152398e73e6189cfa76a5a429391d2/data-collector-3.4.0-sources.jar

Caused by: java.lang.IllegalStateException: java.nio.file.FileAlreadyExistsException: /<repo>/.okbuck/ext/<module-path>/data-collector-3.4.0-sources.jar
        at com.uber.okbuck.core.util.FileUtil.symlink(FileUtil.java:124)
        at com.uber.okbuck.core.manager.DependencyManager.lambda$createSymlinks$22(DependencyManager.java:422)
        at com.uber.okbuck.core.manager.DependencyManager.lambda$createSymlinks$23(DependencyManager.java:420)
        at com.google.common.collect.ImmutableList.forEach(ImmutableList.java:407)
        at com.uber.okbuck.core.manager.DependencyManager.createSymlinks(DependencyManager.java:412)
        at com.uber.okbuck.core.manager.DependencyManager.lambda$processDependencies$20(DependencyManager.java:393)
        at com.uber.okbuck.core.manager.DependencyManager.processDependencies(DependencyManager.java:352)
        at com.uber.okbuck.core.manager.DependencyManager.finalizeDependencies(DependencyManager.java:102)
        at com.uber.okbuck.OkBuckGradlePlugin.lambda$apply$0(OkBuckGradlePlugin.java:192)
```

### Setup:
okbuck: `0.48.7`
AGP: `3.3.2`
Gradle: `5.1.1`

**okbuck external dependency configuration:**
```
okbuck {
  externalDependencies {
    downloadInBuck = false
    autoValueConfigurations = [ ...]
  }
  ...
}
```

**`moduleA/build.gradle`**
```
dependenices {
  api 'com.paypal.android.sdk:paypal-one-touch:3.4.0'
  ...
}
...
```

**Steps to reproduce:**
`.\buckw root`

kageiit:
This should not longer be an issue with the latest okbuck. Please re-open if it happens on latest

PR #874, opened 2019-08-28
Account for additional artifact api transform folders in gradle 5.1.1+

anoopc1:
**Description**:
`okbuck` skips artifacts coming from gradle artifact api transform folders (e.g. 
`~/.gradle/caches/transforms-1`)

`Gradle 5.1.1+` creates additional transform folders (`transform-2`) which should be skipped as well.

**Related issue(s)**: https://github.com/uber/okbuck/issues/873


raviagarwal7:
this was merged in https://github.com/uber/okbuck/pull/916

thanks for the PR @anoopc1

PR #882, opened 2019-10-01
[okbuck] Reverse assumption on standard naming of android_modules

andreasnomikos:
**Description**:
Reverse the assumption on naming conventions of android modules for auto-including 
android_resource rules. After this change only rules defined as src_release will have the
res_release automatically included as deps.

**Related issue(s)**:

### DIFF
--- a/buildSrc/src/main/rocker/com/uber/okbuck/template/config/OkbuckAndroidModules.rocker.raw
+++ b/buildSrc/src/main/rocker/com/uber/okbuck/template/config/OkbuckAndroidModules.rocker.raw
@@ -16,7 +16,7 @@ def src_to_res(str):
 def get_libs_res_deps(src_deps):
     mapped_res_deps = []
     for dep in src_deps:
-        if (':src_' in dep and not dep.endswith(':src_main')):
+        if (dep.endswith(':src_release')):
             mapped_res_deps.append(src_to_res(dep))
     return mapped_res_deps
 


buildSrc/src/main/rocker/com/uber/okbuck/template/config/OkbuckAndroidModules.rocker.raw
```diff
@@ -16,7 +16,7 @@ def src_to_res(str):
 def get_libs_res_deps(src_deps):
     mapped_res_deps = []
     for dep in src_deps:
-        if (':src_' in dep and not dep.endswith(':src_main')):
+        if (dep.endswith(':src_release')):
```
kageiit: This would mean any android flavors will not have resources included automatically. I dont think we would want such a breaking change. Can you explain your motivation here?

This can be better done as a config in okbuck extension to control what gets output from the template for specific usecases rather than as the default option for everyone.

andreasnomikos: what are all the "reserved" names of Buck rules you generate for android modules? This filter should only consider those and leave all other names open. Considering all libraries as android libraries (apart from the ones that are named src_main is overly permissive)

kageiit: There is no such list. The names are based on build types and flavors so they can be anything

kageiit: If there are no custom build types and flavors, you usually get just debug and release source sets, but that is not the case for all codebases

andreasnomikos: I think I have a better way around this change for what I am trying to do. (which is to add buck specific rule extentions to the okbuck rules) since this check looks for rule names that start with :src_ I can use the section before the src_ in the name to indicate the rule_type.
will abandon for now

PR #884, opened 2019-10-11
Script to create tinyjavaxtools.jar, for supporting Error Prone in Buck builds

msridhar:
This script goes along with these instructions for enabling Error Prone in Buck builds with OkBuck:

https://github.com/msridhar/okbuck/wiki/Using-Error-Prone-with-Buck-and-OkBuck

The script creates the `tinyjavaxtools.jar` file necessary to make Error Prone work with Buck.  The instructions above should be copied to the OkBuck wiki alongside merging this PR.

I'd like feedback on both the script and the instructions.  In particular, regarding the instructions:
* How can I tweak them to also work for Android builds (I don't think `compileJava` exists there)?
* Is there an easy-ish way to make the instructions compatible with using the Gradle Error Prone  plugin?  Or, absent that, is there at least a way to make them not break the Gradle build?  Like can we set things up so the `compileJava` arguments are only added when an OkBuck task is running?

### DIFF
--- a/tooling/errorprone/createtinyjavaxtools.sh
+++ b/tooling/errorprone/createtinyjavaxtools.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+
+set -ex
+
+DIR=$(pwd)
+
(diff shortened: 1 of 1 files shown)


msridhar:
@kageiit @raviagarwal7 gentle ping

PR #886, opened 2019-11-01
Populate provided deps only if the versionless artifact is not present in runtimeclasspath

raviagarwal7:


raviagarwal7:
@kageiit -- how to enable back the circle ci builds? 

```
This job has been blocked because resource-class:xlarge is not available on your plan. Please upgrade to continue building.

```

kageiit:
> @kageiit -- how to enable back the circle ci builds?
> 
> ```
> This job has been blocked because resource-class:xlarge is not available on your plan. Please upgrade to continue building.
> ```

lets try to downgrade to `resource-class:large`?

I can contact circleci about the xlarge or we can migrate off them if they are made that option unavailable

raviagarwal7:
Downgraded to large but still doesn't work. We might have to move to some other runner

libraries/scalahadooplibrary/build.gradle
```diff
@@ -0,0 +1,11 @@
+apply plugin: "application"
+apply plugin: "java-library"
+
+
+dependencies {
+    // spark is provided by spark-submit
+    compileOnly deps.external.sparkCore
+    compileOnly deps.external.sparkSql
+
+    implementation deps.external.log4j
```
raviagarwal7: @kageiit -- commenting out this line makes `//libraries/scalahadooplibrary/...` to build successfully. 

For some reason, the compile classpath doesn't contain `log4j` which is why the resolved `deps` list doesn't contain the expected log4j dependency, causing the build to fail.  (`intersection(getMain().getExternalDeps(), getProvided().getExternalDeps())`)

On the other hand when `implementation deps.external.log4j` is removed since log4j shows up in the provided list the compilation is successful.

raviagarwal7: was able to figure out the issue. We do set intersection and had to do versionless set intersection instead to get the right output.

kageiit:
circleci is probably related to https://circleci.com/blog/plans-for-optimal-performance-why-circleci-is-changing-our-pricing-model/

Lets try to downgrade to medium to see if we can still run this. Otherwise we will replace it for something else in a followup

kageiit:
```
Hint: Exit code 137 typically means the process is killed because it was running out of memory
Hint: Check if you can optimize the memory usage in your app
Hint: Max memory usage of this container is 4281315328
```

Lets merge without ci if needed. We can fix it after

.idea/codeStyles/Project.xml
```diff
@@ -7,9 +7,7 @@
         <option name="TAB_SIZE" value="2" />
       </value>
     </option>
-    <option name="RIGHT_MARGIN" value="100" />
```
kageiit: Please undo changes to this file

kageiit:
I moved okbuck back over to travis. Feel free to rebase to let ci run checks

PR #911, opened 2020-06-08
Make marking all versions of an artifact as first level configurable

raviagarwal7:
```
externalDependency {
  // is true by default
  markFirstLevelAllVersions = true 
}
```


Atom1985 (approved):
👍👍👍

PR #917, opened 2020-07-16
Add annotations.jar to the kotlin compiler to make it compatible with 1.40 or higher

raviagarwal7:


Atom1985 (commented):
buildSrc/src/main/java/com/uber/okbuck/core/manager/KotlinManager.java

PR #934, opened 2021-01-26
Add ability to configure okbuck buckconfig inclusion

kageiit:
**Description**:
This allows one to configure the inclusion of `.okbuck.buckconfig` to top level `.buckconfig`. The default is to always include it if it is not present


naveenOnarayanan (approved):
LGTM

PR #935, opened 2021-01-26
Make sha256 cache location configurable

kageiit:
**Description**:
This allows configuration of the external dependency sha256 cache of artifacts to be stored outside the project workspace for cache persistence if needed

### DIFF
--- a/buildSrc/src/main/java/com/uber/okbuck/core/manager/DependencyManager.java
+++ b/buildSrc/src/main/java/com/uber/okbuck/core/manager/DependencyManager.java
@@ -75,7 +75,7 @@ public class DependencyManager {
     this.externalDependenciesExtension = okBuckExtension.getExternalDependenciesExtension();
     this.jetifierExtension = okBuckExtension.getJetifierExtension();
     this.buckFileManager = buckFileManager;
-    this.sha256Cache = initSha256Cache(rootProject);
+    this.sha256Cache = initSha256Cache(rootProject, externalDependenciesExtension);
   }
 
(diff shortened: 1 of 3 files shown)


buildSrc/src/main/java/com/uber/okbuck/core/manager/DependencyManager.java
```diff
@@ -515,17 +515,17 @@ private static void computeSha256IfAbsent(File file, HashMap<String, String> sha
...
       return FileUtil.readMapFromJsonFile(projectMappingFile);
     } catch (IOException e) {
       return new HashMap<>();
     }
   }
 
-  private static void persistSha256Cache(Project rootProject, HashMap<String, String> sha256Map) {
-    File projectMappingFile = rootProject.file(OkBuckGradlePlugin.OKBUCK_SHA256);
+  private static void persistSha256Cache(Project rootProject, HashMap<String, String> sha256Map, ExternalDependenciesExtension externalDependenciesExtension) {
+    File projectMappingFile = rootProject.file(externalDependenciesExtension.getSha256Cache());
```
raviagarwal7, line 528: I think this might not work if the specified file is not inside the root project

kageiit: it works. I checked by modifying it in okbuck and verified it was moved outside the project root

kageiit: We need to use aboslute path

raviagarwal7 (approved):

PR #958, opened 2022-02-12
Adding support for Robolectric 4.7.3 and preinstrumented jars

tyvsmith:
Master is currently not in state to release a new version of OKbuck without reverting the work towards upgrading Gradle.

I've created a release branch off of the last tag (v0.52.3) and this is a pull request against that branch. We can cut this release build 0.52.4 from the release branch, then separately patch this diff against master which will require some rebase conflict changes.

This was tested locally against the Uber Android monorepo.

-4.7.3 is needed to address ARM compatibility on m1, as it carries a patch for sqlite4java.
-They also introduced preinstrumented jars which should save 10-12 seconds per test process instantiation.

### DIFF
--- a/buildSrc/src/main/java/com/uber/okbuck/core/manager/RobolectricManager.java
+++ b/buildSrc/src/main/java/com/uber/okbuck/core/manager/RobolectricManager.java
@@ -52,10 +52,13 @@ public final class RobolectricManager {
       apisToDownload = EnumSet.allOf(API.class);
     }
 
+    String preinstrumentedVersion =
+        ProjectUtil.getOkBuckExtension(rootProject).getTestExtension().robolectricPreinstrumentedVersion;
+
     for (API api : apisToDownload) {
       Configuration runtimeApi =
           rootProject.getConfigurations().maybeCreate(ROBOLECTRIC_RUNTIME + "_" + api.name());
-      rootProject.getDependencies().add(runtimeApi.getName(), api.getCoordinates());
+      rootProject.getDependencies().add(runtimeApi.getName(), api.getCoordinates(preinstrumentedVersion));
       runtimeDeps.add(runtimeApi);
     }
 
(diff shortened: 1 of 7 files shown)


buildSrc/src/main/java/com/uber/okbuck/core/manager/RobolectricManager.java
```diff
@@ -115,9 +118,10 @@ public void finalizeDependencies(OkBuckExtension okBuckExtension) {
     API_25("7.1.0_r7", "r1"),
     API_26("8.0.0_r4", "r1"),
     API_27("8.1.0", "4611349"),
-    API_P("P", "4651975"),
```
tyvsmith, line 118: https://github.com/robolectric/robolectric/blob/master/robolectric/src/main/java/org/robolectric/plugins/DefaultSdkProvider.java No longer referenced in robolectric core.

buildSrc/src/main/java/com/uber/okbuck/core/manager/RobolectricManager.java
```diff
@@ -127,11 +131,13 @@ public void finalizeDependencies(OkBuckExtension okBuckExtension) {
       this.frameworkSdkBuildVersion = frameworkSdkBuildVersion;
     }
 
-    String getCoordinates() {
-      return "org.robolectric:android-all:"
+    String getCoordinates(String preinstrumentedVersion) {
+      return "org.robolectric:android-all-instrumented:"
```
raviagarwal7, line 135: we can return the old format and new preinstrumentedVersion based on if `preinstrumentedVersion` is null or not. 
This will keep it backward compatible.

raviagarwal7 (approved):

buildSrc/src/main/java/com/uber/okbuck/core/manager/RobolectricManager.java
tyvsmith: We can, but this is now hardcoded in robolectric 4.7 so there's also a reason to require it and default to i3 (the default version checked in at this time). Thoughts on keeping backwards compatibility with older robo versions instead of a sane default?

raviagarwal7: sounds good then. We can ask users to upgrade to latest robolectric in the release notes.

kageiit (approved):

PR #962, opened 2022-03-09
Fix relative paths for dependencies on arm jdks.

tyvsmith:
These two classes used relative paths, which were failing on zulu jdk.


tyvsmith:
Closing this in favor of merging to master since gradle 7 upgrade is needed anyway.

PR #968, opened 2022-03-14
ManifestMerger was still using relative paths which are having jvm problems

tyvsmith:


tyvsmith:
Abandoning this in favor of including in the gradle7 bump

PR #973, opened 2023-02-16
feat: Add support for exporting raw dependencies

hvvikram:
**Description**:  Add support for exporting raw dependencies

plugin can accept the following configuration

```
{
     okbuck {
            ...
           exportDependencies { enabled = false, file = ".okbuck/raw-deps" }
           ...
}
```

Supports exporting raw dependencies in JSON format.

```
./gradlew okbuck
```

Example exported json snippet:

```
[
  {
    "name": "hadoop-aws",
    "version": "2.8.2",
    "group": "org.apache.hadoop",
    "force": false,
    "excludeRules": [
      "javax.servlet",
      "org.mortbay.jetty",
      "slf4j-log4j12"
    ],
    "transitive": true
  },
 
  ....
]
```

**Related issue(s)**:

### DIFF
--- a/buildSrc/src/main/java/com/uber/okbuck/OkBuckGradlePlugin.java
+++ b/buildSrc/src/main/java/com/uber/okbuck/OkBuckGradlePlugin.java
@@ -162,7 +162,7 @@ public class OkBuckGradlePlugin implements Plugin<Project> {
               new AnnotationProcessorCache(rootBuckProject, buckFileManager, processorBuildFile);
 
           // Create Dependency manager
-          dependencyManager = new DependencyManager(rootBuckProject, okbuckExt, buckFileManager, createDependencyExporter(rootBuckProject));
+          dependencyManager = new DependencyManager(rootBuckProject, okbuckExt, buckFileManager, createDependencyExporter(okbuckExt));
 
           // Create Lint Manager
           String lintBuildFile = LINT_BUILD_FOLDER + "/" + okbuckExt.buildFileName;
@@ -386,12 +386,7 @@ public class OkBuckGradlePlugin implements Plugin<Project> {
     }
   }
 
-  @Nullable
-  private static DependencyExporter createDependencyExporter(Project project){
-    boolean exportDependencies = ProjectUtil.getPropertyAsBoolean(project,"exportDependencies");
-    String exportDependenciesFile = ProjectUtil.getPropertyAsString(project, "exportDependenciesFile",
-        DOT_OKBUCK + "/raw-deps");
-
-    return new JsonDependencyExporter(Paths.get(project.getRootDir().getAbsolutePath(), exportDependenciesFile).toString(), exportDependencies);
+  private static DependencyExporter createDependencyExporter(OkBuckExtension okbuckExt){
+    return new JsonDependencyExporter(okbuckExt.getExportDependenciesExtension());
   }
 }
--- a/buildSrc/src/main/java/com/uber/okbuck/OkBuckGradlePlugin.java
+++ b/buildSrc/src/main/java/com/uber/okbuck/OkBuckGradlePlugin.java
@@ -34,24 +34,21 @@ import com.uber.okbuck.generator.BuckFileGenerator;
 import com.uber.okbuck.template.common.ExportFile;
 import com.uber.okbuck.template.core.Rule;
 import com.uber.okbuck.wrapper.BuckWrapperTask;
+import org.gradle.api.Plugin;
+import org.gradle.api.Project;
+import org.gradle.api.Task;
+import org.gradle.api.artifacts.Configuration;
+
 import java.io.File;
 import java.io.FileOutputStream;
 import java.io.IOException;
 import java.io.OutputStream;
-import java.nio.file.Paths;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Map;
 import java.util.Set;
 import java.util.function.Function;
 import java.util.stream.Collectors;
-
-import org.gradle.api.Plugin;
-import org.gradle.api.Project;
-import org.gradle.api.Task;
-import org.gradle.api.artifacts.Configuration;
-
-import javax.annotation.Nullable;
 
 // Dependency Tree
 //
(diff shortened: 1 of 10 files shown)


buildSrc/src/main/java/com/uber/okbuck/core/util/ProjectUtil.java
```diff
@@ -224,4 +224,19 @@ public static boolean canHaveSources(File dependencyFile) {
...
+    return false;
+  }
+
+  @Nullable
+  public static String getPropertyAsString(Project project, String propertyName, String defaultValue){
+    if(project.hasProperty(propertyName)){
+      return (String)project.getProperties().get(propertyName);
+    }
+    return defaultValue;
+  }
```
raviagarwal7, line 241: we should use OkBuckExtension to define these properties example: https://github.com/uber/okbuck/blob/master/buildSrc/src/main/java/com/uber/okbuck/extension/OkBuckExtension.java#L27

and have it false by default.

buildSrc/src/main/java/com/uber/okbuck/OkBuckGradlePlugin.java
```diff
@@ -379,4 +385,13 @@ private void writeExportedFileRules(Project rootBuckProject, OkBuckExtension okB
       }
     }
   }
+
+  @Nullable
+  private static DependencyExporter createDependencyExporter(Project project){
+    boolean exportDependencies = ProjectUtil.getPropertyAsBoolean(project,"exportDependencies");
+    String exportDependenciesFile = ProjectUtil.getPropertyAsString(project, "exportDependenciesFile",
+        DOT_OKBUCK + "/raw-deps");
```
raviagarwal7, line 393: we can move the default path to the OkBuckExtension itself

buildSrc/src/main/java/com/uber/okbuck/extension/OkBuckExtension.java
```diff
@@ -104,6 +104,7 @@ public class OkBuckExtension {
   public String buckBinaryJava11 =
       DEFAULT_BUCK_BINARY_REPO + ":" + DEFAULT_BUCK_BINARY_SHA + ":java11@pex";
 
+  @Internal private Project project;
```
raviagarwal7, line 107: probably don't need this anymore

buildSrc/src/main/java/com/uber/okbuck/extension/OkBuckExtension.java
```diff
@@ -1,6 +1,7 @@
 package com.uber.okbuck.extension;
 
 import java.io.File;
+import java.nio.file.Paths;
```
raviagarwal7, line 4: remove unused

buildSrc/src/main/java/com/uber/okbuck/core/manager/DependencyManager.java
```diff
@@ -98,6 +104,8 @@ public void resolveCurrentRawDeps() {
       return;
     }
 
+    dependencyExporter.export(rawDependencies);
```
raviagarwal7, line 107: we can move this to the start of `finalizeDependencies` which is responsible to generate all the output.

buildSrc/src/main/java/com/uber/okbuck/core/dependency/exporter/DependencyExporterModel.java
```diff
@@ -0,0 +1,73 @@
...
+  public DependencyExporterModel(ExternalDependency externalDependency) {
+    name = externalDependency.getName();
+    version = externalDependency.getVersion();
+    group = externalDependency.getGroup();
+    force = externalDependency.isForce();
+    transitive = externalDependency.isTransitive();
+
+    if (externalDependency.getExcludeRules() != null) {
+      excludeRules = externalDependency.getExcludeRules().stream().flatMap(e -> {
+        Set<String> set = new TreeSet<>();
```
raviagarwal7, line 33: seems like we will always have one entry, we could use a map instead of flatmap and string instead of set

buildSrc/src/main/java/com/uber/okbuck/core/dependency/exporter/DependencyExporterModel.java
```diff
@@ -0,0 +1,73 @@
...
+  @Nullable
+  private final String name;
+  @Nullable
+  private final String version;
+  @Nullable
+  private final String group;
+  private final boolean force;
+  @Nullable
+  private Set<String> excludeRules = new TreeSet<>();
+  private final boolean transitive;
```
raviagarwal7, line 22: we currently use this property. do we want to export it?

buildSrc/src/main/java/com/uber/okbuck/core/dependency/exporter/ExporterException.java
```diff
@@ -0,0 +1,10 @@
+package com.uber.okbuck.core.dependency.exporter;
+
+import java.io.IOException;
+
+public class ExporterException extends RuntimeException {
```
raviagarwal7, line 5: could extend from `IOException` and can remove the default constructor

raviagarwal7 (commented):
Could we reformat the added/modified files with google-java-formatter

buildSrc/src/main/java/com/uber/okbuck/core/dependency/exporter/DependencyExporterModel.java
hvvikram: removed it

buildSrc/src/main/java/com/uber/okbuck/core/manager/DependencyManager.java
hvvikram: moved to finalizeDependencies

buildSrc/src/main/java/com/uber/okbuck/core/dependency/exporter/ExporterException.java
hvvikram: I switched to accept an Exception in the constructore. IOException is a checked exception, extending that will force the caller to handle the exception.

raviagarwal7 (approved):

raviagarwal7 (approved):

Issue #975, opened 2023-02-24
Okbuck fails if exportDependencies > file path doesn't exist


hvvikram:
Example config

```
exportDependencies {
        enabled = true
        file = "build/raw-deps.json"
}
```

The folder build do not exist. okbuck throws the following exception

```
FAILURE: Build failed with an exception.
* What went wrong:
Execution failed for task ':okbuck'.
> java.nio.file.NoSuchFileException: /home/user/my_project/build/raw-deps.json
```

PR #976, opened 2023-02-24
Create parent directories if it doesn't exists

hvvikram:
**Description**:
Create parent directories for exportDependencies.file path if it doesn't exists.

**Related issue(s)**:
https://github.com/uber/okbuck/issues/975


buildSrc/src/main/java/com/uber/okbuck/core/dependency/exporter/JsonDependencyExporter.java
```diff
@@ -38,10 +48,11 @@ public void export(Set<ExternalDependency> dependencies) {
             .collect(Collectors.toSet());
 
     Gson gson = new GsonBuilder().setPrettyPrinting().create();
+    String file = exportDependenciesExtension.getFile();
```
raviagarwal7, line 51: we could create the Path object here which can be used in `createParent` and `Files.newBufferedWriter`

PR #977, opened 2023-03-17
Add config cleanCacheDir 

hvvikram:
**Description**:
Add config cleanCacheDir which controls whether to delete the cache dir or only the existing dependency rules files, before generating the dependency rule files

Usage
```
okbuck {
   ... 
   externalDependencies {
      ...
      cleanCacheDir = false
   }
   ...
}
```

**Related issue(s)**:

### DIFF
--- a/buildSrc/src/main/java/com/uber/okbuck/core/util/FileUtil.java
+++ b/buildSrc/src/main/java/com/uber/okbuck/core/util/FileUtil.java
@@ -23,6 +23,7 @@ import java.util.HashMap;
 import java.util.stream.Stream;
 import org.apache.commons.io.FileUtils;
 import org.apache.commons.io.filefilter.NameFileFilter;
+import org.apache.commons.io.filefilter.TrueFileFilter;
 import org.gradle.api.Project;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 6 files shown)


buildSrc/src/main/java/com/uber/okbuck/core/util/FileUtil.java
```diff
@@ -156,4 +157,27 @@ public static void persistMapToJsonFile(HashMap<String, String> map, File file)
...
+      if (deleteDir) {
+        try {
+          FileUtils.deleteDirectory(dir);
+        } catch (IOException e) {
+          throw new IllegalStateException("Could not delete dependency directory: " + dir, e);
+        }
+      } else {
+        Collection<File> files = FileUtils.listFiles(dir, new NameFileFilter(fileName), null);
+        for (File file : files) {
+          FileUtils.deleteQuietly(file);
```
raviagarwal7, line 174: would we be able to delete in parallel using parallelstream?

buildSrc/src/main/java/com/uber/okbuck/extension/ExternalDependenciesExtension.java
```diff
@@ -16,6 +16,12 @@ public class ExternalDependenciesExtension {
   /** Specifies the folder where all external dependency rules gets generated. */
   @Input private String cache = ".okbuck/ext";
 
+  /**
+   * Specifies to delete the cache dir, before generating dependency rules If set to false, only
+   * deletes the existing dependency rules files.
+   */
+  @Input private boolean cleanCacheDir = true;
```
raviagarwal7, line 23: cc: @tyvsmith this should help in preserving METADATA files in `3rdparty/jvm`

buildSrc/src/main/java/com/uber/okbuck/core/util/FileUtil.java
hvvikram: ah, good point.

raviagarwal7 (approved):

PR #980, opened 2023-10-13
Delete tmp file and .bak file for android lint runs

tyvsmith:
Android Lint has always had a leftover tmp.bak file after it's run which add up over time. After moving this to buck-out's $TMP, this was obvious.


buildSrc/src/main/rocker/com/uber/okbuck/template/config/OkbuckTargets.rocker.raw
```diff
@@ -187,7 +187,9 @@ def okbuck_lint(
     cmd += "--fullpath "
 
     cmd += "{} `dirname {}` >$LOGFILE 2>/dev/null; ".format(" ".join(lint_options), toLocation(manifest))
-    cmd += "rm $CP_FILE; "
+    
+    #Delete tmp file and tmp.bak file
+    cmd += "rm $CP_FILE*; "
```
tyvsmith, line 192: This is needed because sed -i.bak is used above, and I left that alone for the sake of historical decisions on backing it up. I assume it was used to be preserved in the case of debugging failures.

PR #972, opened 2023-01-30
Add missing API versions for Roboelectric

vicmns:
Adding missing APIS 32 and 33 so they can be resolved by the OkBuck build system

**Description**:
Adding missing APIs levels to Roboelectric manager.

PR #983, opened 2023-12-15
Add support for Android SDK 32, 33, 34

gpolak:
**Description**:
- adding support for latest Android SDKs
- bumping test instrumentation to i4 to match production

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,4 @@ buildSrc/bintray.properties
 bazel-*
 .ijwb
 BUILD.bazel
+buildSrc/secring.gpg
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .gradle
 local.properties
 .DS_Store
+
 **build
 captures
 
@@ -16,9 +17,9 @@ buck-out
 **/BUCK
 
 buildSrc/bintray.properties
+buildSrc/secring.gpg
 
 # Bazel
 bazel-*
 .ijwb
 BUILD.bazel
-buildSrc/secring.gpg
--- a/README-zh.md
+++ b/README-zh.md
@@ -9,7 +9,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.uber:okbuck:0.54.1'
+        classpath 'com.uber:okbuck:0.54.2'
     }
 }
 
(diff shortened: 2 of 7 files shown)


tyvsmith (changes requested):
.54.2 was cut already as the key signing rotation (no code changes). Looks like he just missed the changelog, or technically there was no code change 🤷🏻 , so we'll need to use .54.3.

https://central.sonatype.com/artifact/com.uber/okbuck

gpolak:
@tyvsmith bumped the version to 0.54.3

README-zh.md
```diff
@@ -9,7 +9,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.uber:okbuck:0.54.1'
```
tyvsmith, line 12: This is referring to the previously public release. SO this should be .1 or .2 until we do the actual cut of this new code.

This is the normal flow:
Step 1. Land your PR without any version changes (and optional changelog updates). 
Step 2: Create a Release PR (move version numbers in readmes and property files are set) and you tag
Step 3: Cut release to maven central and github
Step  4: Create PR that resets -SNAPSHOT url to get ready for new commits.

Check the prior version git version history for this pattern.

Or if you're working on a repo like ribs, it's all rolled up into the releasing.sh script.

.gitignore
```diff
@@ -16,6 +17,7 @@ buck-out
 **/BUCK
 
 buildSrc/bintray.properties
+buildSrc/secring.gpg
```
gpolak, line 20: Helps with local testing

gpolak:
@tyvsmith, all green now. I didn't update the pre-instrumented version in `TestExtension.java` due to the same failure as the old PR here. I highlighted the underlying issue [in the comment there](https://github.com/uber/okbuck/pull/972/files#diff-19d468a678bb51a6ed0691c61ffa91a68a37d75ea16a307532e0bd741863ad75R21). We can address this in a separate diff.

.gitignore
tyvsmith: Nit: I don't think there should be a dev flow where there's a local secring. The signings creds should live in $HOME/.gradle which is outside the repo, and the secreing typically sits in a separate user path.

This is fine overall, as it's mostly a no-op. Just more curious on the flow that made this needed.

tyvsmith (approved):

PR #988, opened 2024-04-02
Update to Android tools v30.3.0

jbarr21:
Update the Android tools version to allow for using Lint v31.3.1 or higher which is compatible with K2. Repos can choose their lint version, but to use a newer version, they need to use a compatible manifest merger version which had some breaking changes in v30.3.0.

A bump to Kotlin v1.5 was needed due to compatibility with transitive dependencies. Along the way, the sample project needed to be updates such as removing deprecated kotlin-extensions, disabling unused lints, & backporting removed methods.


.buckconfig
```diff
@@ -25,6 +25,10 @@
 [java]
     jar_spool_mode = direct_to_jar
     duplicates_log_level = fine
+    compile_against_abis = false
```
tyvsmith, line 28: Why?

build.gradle
```diff
@@ -1,7 +1,14 @@
 buildscript {
     apply from: rootProject.file("dependencies.gradle")
     repositories {
-        google()
+        maven {
+            url 'https://dl.google.com/android/maven2/'
```
tyvsmith, line 5: doesn't google() exist?

jbarr21: due to rocker and other breaking upgrades since we are on AGP 4.x, we need to set the property of the maven url to use pom instead of gradle metadata & we can't do this w/ `google`

.buckconfig
jbarr21: the plugin works in monorepo but okbuck tests are failing due to ASM in Buck trying to perform the class ABI on a java 17 class from a dependency. i'm working on a better fix

tyvsmith (approved):

PR #989, opened 2025-02-16
Update Robolectric to 4.14.1

MGaetan89:
## Description

This commit updates Robolectric to its latest version (4.14.1). It also updates the code in `libraries/robolectric-base` to work with the latest version.

## Changes

- Update Robolectric to 4.14.1 (include the version of preinstrumented).
- Add `libraries/robolectric-base` to the list of Gradle modules to load.
- Add missing jUnit dependency to `libraries/robolectric-base`.
- Replace `BuckRobolectricTestRunner#getAppManifest(Config)` with `BuckRobolectricTestRunner#getManifestFactory(Config)`. The former is no longer public, and the recommended alternative is to use a custom `ManifestFactory`.
- Create `BuckManifestFactory` to provide custom information for the creation of `AndroidManifest`.


MGaetan89:
It looks like the build failed downloading some dependencies. Is it possible to restart it?

MGaetan89:
I've updated my branch with the latest changes from `master`. It should be ready for review now.

MGaetan89:
I think I'll have to update the [Java version](https://github.com/uber/okbuck/blob/6160366a0b356032d47930555f535ed0a7def4dc/.github/workflows/ci.yml#L26) used in the project to make this work.
I'll submit a dedicated PR for that, because it seems to impact other things too.

MGaetan89:
I've updated this PR to just migrate deprecated code from Robolectric with the recommended alternatives.
Upgrading Robolectric itself needs too many additional changes throughout the project.

gpolak:
Thanks @MGaetan89. Looks good, but let me do some local verification before merging. Stay tuned...

And yes, we need to build Buck against JDK 17 before we can move to later versions of Robolectric.

gpolak:
Looks good @MGaetan89. Sorry for the long delay, the JDK 17 migration took longer than anticipated - thanks again!

PR #990, opened 2025-02-18
Update actions used in GitHub Actions workflows

MGaetan89:
**Description**

- Update the runner image to `ubuntu-24.04`. `ubuntu-20.04` is now deprecated: https://github.com/actions/runner-images/issues/11101
- Update `actions/checkout` to v4.
- Update `actions/setup-java` to v4 and set the distribution to `temurin`.

**Note**

- It's possible to use `ubuntu-latest` instead of `ubuntu-24.04`. Let me know which option you prefer.
- `actions/setup-java` supports a [wide range of distributions](https://github.com/actions/setup-java#supported-distributions). Let me know if you'd prefer to use something else than `temurin`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,17 +13,18 @@ jobs:
       SKIP_OKBUCK: true
       EXTRA_OKBUCK_ARGS: "--quiet --stacktrace"
       EXTRA_BUCK_ARGS: "-v 0"
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-24.04
     strategy:
       matrix:
         build_cmd: [build, lint, test]
     steps:
       - name: Checkout
-        uses: actions/checkout@v1
+        uses: actions/checkout@v4
       - name: Install JDK
-        uses: actions/setup-java@v1
+        uses: actions/setup-java@v4
         with:
           java-version: 11
+          distribution: temurin
       - name: Run ${{ matrix.build_cmd }}
         run: ./tooling/ci/build_cmd.sh ${{ matrix.build_cmd }}
       - name: Upload snapshot
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,6 +25,10 @@ jobs:
         with:
           java-version: 11
           distribution: temurin
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.8'
       - name: Run ${{ matrix.build_cmd }}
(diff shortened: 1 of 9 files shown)


tyvsmith:
> It's possible to use ubuntu-latest instead of ubuntu-24.04. Let me know which option you prefer.
24.04 is fine.

> actions/setup-java supports a [wide range of distributions](https://github.com/actions/setup-java#supported-distributions). Let me know if you'd prefer to use something else than temurin.
We use temurin internally, this is fine.

CI seems to be failing resolving some maven deps.

MGaetan89:
The failing dependencies are: `com.tbruyelle.rxpermissions2:rxpermissions:0.9.5` and `com.github.promeg:xlog-android-idle:2.1.1`.

## rxpermissions

Looking on Maven Central, the only result appears to be a fork of that dependency: https://central.sonatype.com/artifact/io.github.mayampi01/com.tbruyelle.rxpermissions2.

If it's ok for you, I can change the coordinate of the dependency to the one I linked above.
But it looks like this dependency may not be used, so maybe I could simply remove it.
What do you prefer?

## xlog-android-idle

For this one, there doesn't seem to be any result on Maven Central: https://central.sonatype.com/search?q=xlog-android-idle.
So I'm not sure how to move forward here.

The dependency seems to be used only in the example app, so maybe it could be removed too?

tyvsmith:
I assume both of these went down with jcenter. They both seem overkill for having in a sample app, and I don't see much/any usage.

For rxPermissions, I see none, let's kill it.

Xlog looks fine to delete too, but need the couple references stripped. Was used to support annotation based logging, I don't see any more uses of that in the app.
https://github.com/promeG/XLog

@raviagarwal7 Do you have an idea about these old deps?

raviagarwal7:
Agreed, we can remove both the deps and cleanup usages if any. The deps are used in sample apps which can be simplified.

MGaetan89:
Ok, I'll update the PR accordingly soon.

MGaetan89:
I've updated my PR to use Python 3.8 in the `ci.yml` workflow. I've tested it on my fork, and it seems to work fine now.

A couple of notes:
- Python 3.8 has reached [end of life](https://devguide.python.org/versions/). Maybe the `buckw` script should be updated to support more recent versions.
- The `fossa.yml` workflow uses `fossas/fossa-cli` v1, which is deprecated. It is recommended to use v3.

MGaetan89:
And now it works 🎉

tyvsmith:
Looks good, thanks for the follow-up and helping get this fixed!

PR #993, opened 2025-09-29
Add support for setting 'labels' in prebuilt rules

ckilian867:
In this commit, we add support for setting the 'labels' field to the template for the prebuilt macros. This allows for setting specific lables on dependencies, which can be used for validation and querying logic.

There are situations where we may want to make decisions on specific sets of dependencies with a macro, or even make specific sets of dependencies easier to query. 'labels' is already Buck's way of adding arbitrary metadata to a rule and can be leveraged here. 

**Related issue(s)**:
None

### DIFF
--- a/Usage.md
+++ b/Usage.md
@@ -61,10 +61,20 @@ okbuck {
     experimental {
         transform = true
     }
-    
+
     externalDependencies {
         cache = "3rdparty/jvm"
         cleanCacheDir = true
+        labelsMap = [
+            "com.example:library:1.0.0": [
+                "category=utility",
+                "license=apache-2.0"
+            ],
+            "junit:junit:4.13.2": [
+                "category=testing",
+                "test_framework=true"
+            ]
+        ]
     }
 }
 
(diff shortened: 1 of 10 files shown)


buildSrc/src/main/java/com/uber/okbuck/composer/java/PrebuiltRuleComposer.java
```diff
@@ -13,19 +13,41 @@
...
+    if (labelsMap == null || labelsMap.isEmpty()) {
+      return ImmutableSet.of();
+    }
+
+    // Use Maven coordinates as key
+    String key = dependency.getMavenCoordsForValidation();
+    List<String> labels = labelsMap.get(key);
+
+    // Return empty set if labels is null or empty
+    return (labels == null || labels.isEmpty()) ? ImmutableSet.of() : ImmutableSet.copyOf(labels);
```
tyvsmith, line 36: How about simplifying this:

```
  List<String> labels = (labelsMap == null)
      ? Collections.emptyList()
      : labelsMap.getOrDefault(dependency.getMavenCoordsForValidation(), Collections.emptyList());
  return ImmutableSet.copyOf(labels);
```

Usage.md
```diff
@@ -85,6 +95,10 @@ please read the [Exopackage wiki](https://github.com/uber/okbuck/wiki/Exopackage
 +  `extraBuckOpts` provides a hook to add additional configuration options for buck [android_binary](https://buckbuild.com/rule/android_binary.html) rules
 +  `wrapper` is used to configure creation of the buck wrapper script.
  - `repo` - The git url of any custom buck fork. Default is none.
++  `externalDependencies` block configures external dependency resolution and generation:
++ - `cache` - Specifies the folder where external dependency rules are generated. Default is `.okbuck/ext`
++ - `cleanCacheDir` - Whether to delete the cache directory before generating dependency rules. Default is `true`
++ - `labelsMap` - Map of dependency coordinates to labels for prebuilt rules. Keys are Maven coordinates in format `"groupId:artifactId:version"`, values are lists of arbitrary strings
```
tyvsmith, line 101: An example usecase may be helpful.

tyvsmith (approved):

